PR #1, opened 2021-09-15
expose init function in ts definition. use instance props over fields

kellyo-cricut:
Tested locally, regenerated as expected in CriGeo bindings.  Are there any tests that need to be ran/adjusted/added?

### DIFF
--- a/Sources/FishyJoesRuntime/js/wasm-napi.js
+++ b/Sources/FishyJoesRuntime/js/wasm-napi.js
@@ -70,29 +70,32 @@ const kNoException = Symbol('kNoException');
 const hasOwnProperty = Function.prototype.call.bind(Object.prototype.hasOwnProperty);
 
 export class NAPI {
-  indirectFunctionTable = undefined;
-  memory = undefined;
-  view = undefined;
-  instanceData = { data: 0 };
+  constructor() {
+    this.indirectFunctionTable = undefined;
+    this.memory = undefined;
+    this.view = undefined;
+    this.instanceData = { data: 0 };
 
-  scopes = [];
-  handles = [null];
-  references = [];
+    this.scopes = [];
+    this.handles = [null];
+    this.references = [];
 
-  externalData = new WeakMap();
-  wrapData = new WeakMap();
-  typeTags = new WeakMap();
-  finalizationRegistry = new FinalizationRegistry((item) => {
-    const [cbIdx, envPtr, data, hint] = item
-    try {
-      this.indirectFunctionTable.get(cbIdx)(envPtr, data, hint);
-    } catch (e) {
-      internalBinding('task_queue').triggerFatalException(e);
-    }
-  });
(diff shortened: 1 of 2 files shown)


Sources/FishyJoesRuntime/js/wasm-napi.js
```diff
@@ -258,478 +261,479 @@ export class NAPI {
...
+                                           const dataView = this.load(dataviewIdx);
+                                           if (!util.types.isDataView(dataView)) {
+                                             return NAPI_INVALID_ARG;
+                                           }
+                                           this.view.setUint32(byteLengthPtr, dataView.byteLength);
+                                           this.view.setUint32(dataPtr, 0);
+                                           this.writeValue(arraybufferPtr, dataView.buffer);
+                                           this.view.setUint32(byteOffsetPtr, dataView.byteOffset);
+                                           return NAPI_OK;
+                                         }),
```
cobbal: Can we line break differently here so that huge indent doesn't happen?

Sources/FishyJoesRuntime/js/wasm-napi.js
```diff
@@ -258,478 +261,479 @@ export class NAPI {
...
+                                                  this.view.setBigUint64(wordsPtr + (i * 64), bv, true);
+                                                  ull >>= 64n;
+                                                }
+                                                while (ull > 0) {
+                                                  i += 1;
+                                                  ull >>= 64n;
+                                                }
+                                                this.view.setUint32(wordCountPtr, i, true);
+                                                return NAPI_OK;
+                                              }),
```
cobbal: and here

Sources/FishyJoesRuntime/js/wasm-napi.js
```diff
@@ -258,478 +261,479 @@ export class NAPI {
...
+                                             Int32Array,
+                                             Uint32Array,
+                                             Float32Array,
+                                             Float64Array,
+                                             BigInt64Array,
+                                             BigUint64Array,
+                                           ][type](this.load(arraybufferIdx), byteOffset, length);
+                                           this.writeValue(resultPtr, this.store(ab));
+                                           return NAPI_OK;
+                                         }),
```
cobbal: and here

cobbal (changes requested):

Sources/FishyJoesRuntime/js/wasm-napi.js
kellyo-cricut: whoops! yeah, i think this may have been auto-formatted by the IDE when i dropped it in here

kellyo-cricut: the file as-is i just lifted from this commit: 6bb1874bc817facf812a3fd2a3a17183a2e4dbf0
we use [prettier](https://prettier.io/) on desktop for formatting- should i run this file through that, or just adjust these few spots by hand?

PR #4, opened 2021-10-18
Use node's own NAPI tests to find a lot of problems in the wasm implementation

cobbal:
Most of the important changes are in wasm-napi.js.

Anything in "node-tests/common" or "node-tests/js-native-api" is copied without modification from https://github.com/nodejs/node


Sources/FishyJoesNodeRuntime/js/wasm-napi.js
```diff
@@ -1281,31 +1538,124 @@ export class NAPI {
...
+    this.references[this.nextRef] = ref;
+    this.nextRef += 1;
+    return this.nextRef - 1;
+  }
+
+  allocateErrorMessages() {
+    const tablePtr = this.malloc(4 * error_messages.length);
+    for (let i = 0; i < error_messages.length; i++) {
+      let stringPtr = 0
+      if (error_messages[i] === undefined) {
```
kellyo-cricut: Can we invert this condition instead of having an empty `if` block?

kellyo-cricut (approved):

PR #18, opened 2021-12-03
Foundation data

mstokercricut:
* Added translation for UInt8,16,32,64 Int8,16,32,64 and Float types (where possible)
* Added translation of Foundation.Data
* Resolved name shadowing under certain circumstances
* Moved to `jboolean` instead of `Bool` for methods working with these directly
* Started using fully-qualified type names in Kotlin, to avoid local type shadowing


Sources/FishyJoesCore/FishyJoesContext.swift
```diff
@@ -140,10 +142,20 @@ public class FishyJoesContext {
...
 
         let primitiveTypeMap = [
-            "Double": (c: "double", ts: "number", jni: JNIType.double),
-            "Int": (c: "int", ts: "number", jni: JNIType.long),
+            "Bool": (c: "_Bool", ts: "boolean", jni: JNIType.boolean),
             "UInt8": (c: "uint8_t", ts: "number", jni: JNIType.byte),
-            "Bool": (c: "_Bool", ts:"boolean", jni: JNIType.boolean),
+            "UInt16": (c: "uint16_t", ts: "number", jni: JNIType.int),
+            "UInt32": (c: "uint32_t", ts: "number", jni: JNIType.long),
+            // TODO: JavaScript would sadly be using 64-bit double for 64-bit int, loosing precision, so need BigInt, and java is sad here too
```
cobbal:
```suggestion
            // TODO: JavaScript would sadly be using 64-bit double for 64-bit int, losing precision, so need BigInt, and java is sad here too
```

Sources/FishyJoesCore/TranslatedTypes/TranslatedEnum.swift
```diff
@@ -100,31 +104,31 @@ struct TranslatedEnum: TranslatedType {
             )
         } else {
             fragment.outputBlock("extension \(sourceType.name): FishyJoesNodeRuntime.NodeConverter {") {
-                fragment.outputBlock("public static func fromNode(_ value: napi_value?, env: napi_env) throws -> Self {") {
-                    fragment.output("let instanceData = try FishyJoesNodeRuntime.InstanceData.data(for: env)")
+                fragment.outputBlock("public static func fromNode(_ __value: napi_value?, env __env: napi_env) throws -> Self {") {
```
cobbal: I don't think double underscores are used anywhere else in the library like this. Not that there's only one convention, but adding yet another one is probably a bad idea.

Sources/FishyJoesJavaRuntime/JavaConverter.swift
```diff
@@ -505,13 +679,63 @@ extension Tuple2Converter: JavaConverter where T0: JavaConverter, T1: JavaConver
...
+fileprivate var tuple3FirstMethod: jmethodID!
+fileprivate var tuple3SecondMethod: jmethodID!
+fileprivate var tuple3ThirdMethod: jmethodID!
+
+extension Tuple3Converter: JavaConverter where T0: JavaConverter, T1: JavaConverter, T2: JavaConverter {
+    public static var javaClass: jclass? {
+        tuple4Class
+    }
+    public static var javaDescriptor: String {
+        "Lcom/cricut/fishyjoes/runtime/Tuple3;"
```
cobbal: This should use kotlin's Triple class instead

cobbal (changes requested):

cobbal (approved):
Awesome.

PR #20, opened 2021-12-08
Added an integration test

cobbal:


mstokercricut (approved):
Node tests could use work, but this looks good for Kotlin.

PR #23, opened 2021-12-09
Unsigned integers

mstokercricut:


Sources/FishyJoesCore/Unparse/KotlinClass.swift
```diff
@@ -143,11 +144,37 @@ class KotlinClass {
...
+        switch self {
+        case let .unsigned(jniType): return "U\(jniType.valueType)"
+        case let .optional(wrapped): return "\(wrapped.kotlinType)?"
+        default: return jvmType
+        }
+    }
+    
+    var toJVMType: String {
+        switch self {
+        case .unsigned: return ".to\(jvmType)()"
```
cobbal, line 174: I think an optional case is also needed here and in `toKotlinType`

integration-tests/TestAPI-bindings/kotlin/local.properties
```diff
@@ -0,0 +1,8 @@
+## This file must *NOT* be checked into Version Control Systems,
```
cobbal: Pretty sure this shouldn't be checked into Version Control Systems

kotlin-runtime/local.properties
```diff
@@ -0,0 +1,8 @@
+## This file must *NOT* be checked into Version Control Systems,
```
cobbal: ditto

cobbal (changes requested):

Sources/FishyJoesCore/Unparse/KotlinClass.swift
```diff
@@ -168,12 +168,14 @@ extension KotlinClass.KType: CustomStringConvertible {
     var toJVMType: String {
         switch self {
         case .unsigned: return ".to\(jvmType)()"
+        case let .optional(wrapped): return "?.to\(wrapped.jvmType)()"
```
cobbal: This will probably break non-unsigned optionals. What about `wrapped.toJvmType.map { "?\($0)" }` and making toJVMType return `String?`

cobbal (changes requested):

cobbal (approved):

cobbal:
We should probably figure out how to not have the integration tests depend on the published base package though...

PR #24, opened 2021-12-13
Int64 & UInt64

mstokercricut:
* Fixed a conversion bug with Int64 and UInt64. 
* Made other conversions more conformant to one another as well.


Sources/FishyJoesNodeRuntime/NodeConverter.swift
```diff
@@ -79,25 +79,26 @@ extension UInt32: NodeConverter {
...
-//        try check(napi_create_uint64(env, UInt32(value), &result))
-//        return result
-//    }
-//}
+extension UInt64: NodeConverter {
+    public static func fromNode(_ value: napi_value?, env: napi_env) throws -> Self {
+        var result = UInt64(0)
+        var lossless = true
+        try check(napi_get_value_bigint_uint64(env, value, &result, &lossless))
+        if !lossless { throw JSException(message: "requested conversion of invalid bigint to uint64") }
```
cobbal: Not a great error message. Something like "bigint \(...) outside of range of UInt64"?

cobbal (changes requested):
Needs testing with 32 bit wasm and checking that UInt64.maxValue doesn't show up as a negative bigint.

integration-tests/TestAPI-bindings/node-test/test.ts
```diff
@@ -1,6 +1,201 @@
...
+    // Testing Utilities
+    function assertEqual<T>(value: T, expected: T) {
+        console.assert(value == expected, value + " was expected to be " + expected)
     }
+
+    function assertArraysEqual<T>(array: Array<T>, expected: Array<T>) {
+        console.assert(array.length == expected.length, array.every(function(e, i) { return e === expected[i] }), array + " was expected to be " + expected)
+    }
+
+    function assertObjectsEqual<T>(obj, expected) {
```
cobbal: I think `assert.deepStrictEqual` would be preferable here: https://nodejs.org/api/assert.html#assertdeepstrictequalactual-expected-message

integration-tests/TestAPI-bindings/node-test/test.ts
```diff
@@ -1,6 +1,201 @@
 import(`${MODULE_PATH}/TestAPI.js`).then(({ TestAPI }) => {
-    console.log(TestAPI);
-    if (TestAPI.ExportedByReference.create().text !== "Hello, World!") {
-        throw "test failed. Also, figure out how to do TS tests better";
+    // Testing Utilities
+    function assertEqual<T>(value: T, expected: T) {
```
cobbal: `assert.equal`

integration-tests/TestAPI-bindings/node-test/test.ts
```diff
@@ -1,6 +1,201 @@
...
+        console.assert(value == expected, value + " was expected to be " + expected)
     }
+
+    function assertArraysEqual<T>(array: Array<T>, expected: Array<T>) {
```
cobbal: `assert.deepStrictEqual`

cobbal (changes requested):
```
Testing Primitive Values...
Assertion failed: -1 was expected to be 18446744073709551615
Testing Arrays of Primitive Values...
Testing Arrays of Optional Primitive Values...
Testing Functions Taking and Returning Primitive Types...
Assertion failed: -1 was expected to be 18446744073709551615
Testing Functions Taking and Returning Optional Primitive Types...
Assertion failed: -1 was expected to be 18446744073709551615
Testing Objects with Primitive Members...
```
wasm-napi.js needs to be updated. Also `console.assert` doesn't seem to stop things

Sources/FishyJoesNodeRuntime/js/wasm-napi.js
```diff
@@ -551,11 +551,11 @@ export class NAPI {
         return NAPI_OK;
       }),
       napi_create_bigint_int64: this.wrap((envPtr, value, resultPtr) => {
-        this.writeU32(resultPtr, this.store(value));
+        this.writeU32(resultPtr, this.store(BigInt.asIntN(64, value)));
```
cobbal: Don't think this one is needed, but I'd expect it to be used in the other uint64 function

cobbal (changes requested):

cobbal (approved):

PR #28, opened 2022-01-10
refactor: Clean up and centralize swift reference type code into runtime

cobbal:
Moving kotlin's swiftReference fields to a common superclass allows for code reuse between function types, generated reference types, and in the future protocol types.


mstokercricut (approved):
This universally simplifies things, so I'm all for it. Any foreseeable performance issues?

PR #29, opened 2022-01-31
Kotlin protocols

cobbal:


Sources/FishyJoesCore/FishyJoesContext.swift
```diff
@@ -43,21 +43,23 @@ public class FishyJoesContext {
         var collectedFragments: [SourceFragment] = []
 
         // Collect type information before starting translation
-        for translatedType in templateContext.types.all.compactMap(translate(typeDefinition:)) {
+        for translatedType in templateContext.types.types.compactMap(translate(typeDefinition:)) {
```
mstokercricut, line 46: Thats... confusing.

Sources/FishyJoesCore/FishyJoesContext.swift
```diff
@@ -43,21 +43,23 @@ public class FishyJoesContext {
...
         // Translate
-        for type in templateContext.types.all + templateContext.types.extensions {
-            for method in type.allMethods.compactMap(Method.init) {
-                collectedFragments.append(contentsOf: kotlinTranslator.translate(method: method, context: self))
-            }
-            for variable in type.rawVariables {
-                collectedFragments.append(contentsOf: kotlinTranslator.translate(variable: variable, context: self))
-            }
+        let allMethods = Set(templateContext.types.types.flatMap { $0.allMethods.compactMap(Method.init) })
+        for method in allMethods.sorted(by: { $0.callName < $1.callName }) {
```
mstokercricut, line 54: Sorting will be nice.

Sources/FishyJoesCore/KotlinTranslate.swift
```diff
@@ -96,16 +96,17 @@ class KotlinTranslate {
         let jvmGetName = "__jni_\(exportAnnotation.kind == .asMethod ? "" : "get_")\(kotlinName)"
         let jvmSetName = "__jni_set_\(kotlinName)"
 
-        let selfExpression: String
+        var selfExpression: String
```
mstokercricut, line 99: This could still be a `let` right? Just gets its value in 3 different places, but the compiler can ensure that happens.

Sources/FishyJoesCore/KotlinTranslate.swift
```diff
@@ -283,4 +293,71 @@ class KotlinTranslate {
             )
         }
     }
+
+    func conform(_ kotlinClass: KotlinClass, to protoName: String, context: FishyJoesContext) {
```
mstokercricut, line 297: This is nice and clear.

Sources/FishyJoesJavaRuntime/JavaConverter.swift
```diff
@@ -9,6 +9,9 @@ public protocol JavaConverter: Converter {
 
     static var javaClass: jclass? { get }
 
+    // only defined for protocols
+    static var externalWitnessClass: jclass? { get }
```
mstokercricut, line 13: Seems odd to define at this level. Just for convenience?

integration-tests/TestAPI/Sources/TestAPI/AProtocol.swift
```diff
@@ -0,0 +1,30 @@
...
+
+/// <!-- FishyJoes.export(AProtocolImplementation) -->
+public struct AProtocolImplementation: AProtocol {
+    public var foo: String
+
+    public init(foo: String) {
+        self.foo = foo
+    }
+
+    // TODO: should we infer the export from protocol requirements?
```
mstokercricut, line 25: A good question... how hard is that?

Sources/FishyJoesCore/Method.swift
```diff
@@ -11,11 +11,12 @@ struct Method {
 
     let documentation: [String]
 
-    let definedIn: BetterType?
+    var definedIn: BetterType?
```
mstokercricut, line 14: Having this be mutable seems scary in terms of maintenance of FishyJoes. Any other way?

Sources/FishyJoesCore/Unparse/KotlinClass.swift
```diff
@@ -12,35 +12,41 @@ class KotlinClass {
...
 
     struct Method {
-        let documentation: [String]
-        let isStatic: Bool
-        let isOverride: Bool
-        let name: String
-        let parameters: [(labelComment: String?, name: String, type: KType)]
-        let returnType: KType
-        let body: String?
+        var documentation: [String]
```
mstokercricut, line 15: More scary mutability...

Sources/FishyJoesCore/FishyJoesContext.swift
mstokercricut: Any other places sorting would make the generated files easier to navigate? The reduction in PR changes due to rearranging would be helpful too...

mstokercricut (approved):
No blockers here. Could be a bit clearer in some areas, but this certainly improves things in general. Commenting of large logical blocks would make the code easier to follow, especially during unparse.

Sources/FishyJoesCore/FishyJoesContext.swift
cobbal: Most output is deterministic. Any that isn't is a bug.

Sources/FishyJoesJavaRuntime/JavaConverter.swift
cobbal: No. It comes down to which classes JNI methods need to be inserted into. I should document that better.

Sources/FishyJoesCore/Method.swift
cobbal: A mutable struct isn't that scary, since copies will be made anyway. This just allows for easier copy-but-with-slight-modification than calling the initializer with almost all properties the same.

Sources/FishyJoesCore/FishyJoesContext.swift
dvosscricut: types.types (which includes protocol types)? it's so clear to me what that means versus types.all (which excludes protocol types) :o)

Sources/FishyJoesCore/Unparse/KotlinClass.swift
dvosscricut: whooooooooo! 👻

dvosscricut:
Done by https://github.com/cricut/FishyJoes/pull/173

PR #30, opened 2022-04-01
Upgrade to swift-wasm 5.6, add wasm-opt to build process if available

cobbal:
I probably want to factor out the version number at some point, but it's not easy to get it into all the right places...


mstokercricut (approved):
LGTM

PR #32, opened 2022-04-11
Swift package-dump

mstokercricut:
* Responded to change in format of swift package-dump.
* Updated to Sourcery 1.8.0


cobbal (approved):
Approved if CI passes

PR #38, opened 2022-04-20
Lots more dynamic libraries

cobbal:
java runtime is a dynamic lib, all modules that FishyJoes runs against should be dynamic libs when SWIFT_PACKAGE_FORCE_DYNAMIC=1


Sources/FishyJoesCore/FishyJoesContext.swift
```diff
@@ -43,7 +40,10 @@ public class FishyJoesContext {
             fatalErr("must provide `requiredModules` as argument to sourcery")
         }
         self.templateContext = context
-        self.module = module
+        self.module = Module(
+            name: module,
+            dependencies: requiredModulePaths.map { (($0 as NSString).lastPathComponent as NSString).deletingPathExtension }
```
bscothern-Cricut, line 45: Since Sorcery only works on macOS maybe it isn't a big deal but in general we should probably standardize on a Path type for FishyJoes and Popplers. There are libraries dedicated to just Paths or there is also the Swift System library from apple which supports stuff like that and other common OS operations and things like file IO. Then we can use them to get the last component and make sure we are consistent and correct regardless of the platform it is being run on. That said, NSString is probably good enough for now even if the OSS one isn't correct (it might be but I don't know if it is or not).

bscothern-Cricut (approved):
Looks good

mstokercricut (approved):
LGTM

PR #40, opened 2022-04-26
add swiftlint

cobbal:


Sources/FishyJoesExecute/CodeGen.swift
```diff
@@ -19,25 +19,25 @@ let dylibExt: String = {
 
 struct CodeGen: ParsableCommand {
     @Flag(name: .shortAndLong, help: "suppress verbose output")
-    var quiet: Bool = false
+    var quiet = false
```
bscothern-Cricut, line 22: I think this is a rule I have started turning off in my projects for this exact reason. I have had too many weird compilation issues/errors when not specifying the type with property wrappers. I'm fine either way though.

integration-tests/TestAPI-bindings/Sources/Generated/JavaInterface/Strings+java.swift
```diff
@@ -1,5 +1,6 @@
 // Generated using Sourcery 1.8.0 — https://github.com/krzysztofzablocki/Sourcery
 // DO NOT EDIT
+// swiftlint:disable superfluous_disable_command unused_closure_parameter syntactic_sugar
```
bscothern-Cricut, line 3: We can also change the paths that `SwiftLint` looks at to not include the generated sources if you would rather do that.

bscothern-Cricut (approved):

integration-tests/TestAPI-bindings/Sources/Generated/JavaInterface/Strings+java.swift
cobbal: Thought about that, but I actually want the generated code to look right.

PR #41, opened 2022-04-27
Fix SwiftPackage parsing on swift 5.5 with tests

cobbal:


mstokercricut (approved):
Good tests. Better treatment of the package data structure is good too.

PR #50, opened 2022-05-20
Cobbal/critext wasm

cobbal:
Typescript is now able to export types for dependent modules as well


Sources/FishyJoesCore/Utilities.swift
```diff
@@ -35,3 +35,12 @@ extension Array {
         (self[safe: 0], self[safe: 1], self[safe: 2], self[safe: 3])
     }
 }
+
+extension Result where Failure == Never {
```
mstokercricut, line 39: This... seems like it should not need helping. Odd this case it not covered by the standard library when the failure type is known as Never at compile time...

mstokercricut (approved):

PR #51, opened 2022-05-26
Cobbal/sonarcloud

cobbal:
Adds coverage for swift/kotlin on integration, unit, and any other tests I could find.

Coverage not yet reported:
 - wasm (not supported)
 - js (coverage generated, but not reported because it's temporary files)


mstokercricut (approved):
Looks good. An example of data to be parsed in lcov-to-xml.sh might help the reader, though.

PR #59, opened 2022-06-09
Cobbal/ts testing

cobbal:


integration-tests/TestAPI-bindings/node-test/TestAPI.test.ts
```diff
@@ -0,0 +1,367 @@
+import { TestAPI } from 'TestAPI';
```
mstokercricut: Now that this is not within that strange context, can these tests live in their own files? Each `// FooTests` comment indicates where test cases are broken (in the Kotlin tests, at least).

mstokercricut (changes requested):

integration-tests/TestAPI-bindings/node-test/TestAPI.test.ts
cobbal: They certainly can. Will fix tomorrow.

kellyo-cricut: another common approach to grouping these is using `describe` as a wrapper for a function and/or class, and having `it`/`test` blocks inside each.

```
describe('primitive values', () => {
  it('should correctly map numeric types', () => {
    expect(...)
  })
})

kellyo-cricut (approved):

mstokercricut (approved):

PR #62, opened 2022-06-28
Extract stdlib from jar on non-android linux

cobbal:


kotlin-runtime/src/main/kotlin/com/cricut/fishyjoes/runtime/LibraryLoader.kt
```diff
@@ -51,11 +50,14 @@ object LibraryLoader {
         } else {
             error("$libraryName unsupported OS: $osName")
         }
+        return extractLibraryFromJar(jarPath, "$prefix$libraryName.$ext")
+    }
 
+    private fun extractLibraryFromJar(jarPath: String, outFileName: String): File {
```
steelenelson: This method name is confusing to me when I read jarPath I think path to a jar but reading the code I think you mean path inside of the current jar?

steelenelson (approved):

PR #64, opened 2022-07-11
c hashtag

cobbal:
Things compile, but are still too broken to run.


Sources/FishyJoesCommonRuntime/Default.swift
```diff
@@ -0,0 +1,18 @@
+public protocol Default {
```
bscothern-Cricut: Should this be `Defaultable` instead of `Default`?

bscothern-Cricut (approved):
Looking good!

PR #65, opened 2022-07-22
More C#

cobbal:


Sources/FishyJoesCSharpRuntime/CSharpConverter.swift
```diff
@@ -59,676 +84,410 @@ extension VoidConverter: CSharpConverter {
...
+    public static func toCSharp(_ value: Self) throws -> Self { value }
+
+    public static func fromCSharp(object: csObject) throws -> Self {
+        try Env.check { exn in try Env.unwrap(valueMethod)(object, exn) }
+    }
+    public static func toCSharpObject(_ value: Self) throws -> csObject {
+        try Env.unwrap(constructor)(value)
+    }
+    static var valueMethod: ValueMethod?
+    static var constructor: Constructor?
```
bscothern-Cricut, line 346: Since this is pretty much the same for every time conforming to `PrimitiveCSharpConverter` should they become customization points instead?

Sources/FishyJoesCSharpRuntime/CSharpFunction.swift
```diff
@@ -0,0 +1,519 @@
...
+              let identifier = Env.typeIDsByID[typeID]
+        else {
+            fatalError("unregistered typeID \(name)")
+        }
+        SwiftFunctionImpl.constructors[identifier] = constructor
+        SwiftFunctionImpl.invoke6s[identifier] = invoke
+    }
+}
+
+// This code is really repetitive, but I'm not sure how it could be made better...
```
bscothern-Cricut, line 156: Outside of variadic generics or having stuff associated with the AnyFunction# stuff that changes what type `invoke` is to take a tuple and passing in arguments that way I don't know that there is much. We can do to cut down repetition. The tuple and invoke stuff gets messy, I know because I was just doing some experiments with that kind of stuff in a personal library that lakes you take a `(N...) -> R` and turn it into function that take tuples like `((N...)) -> R`.

bscothern-Cricut (approved):

Sources/FishyJoesCSharpRuntime/CSharpConverter.swift
cobbal: Compiler didn't like that solution for some reason. Had to copy/paste instead :(

PR #67, opened 2022-08-04
Fix all C# memory leaks (probably)

cobbal:
bump to dot net 6.0


mstokercricut (approved):
Those single-item structs make things *sooo* much nicer.

PR #69, opened 2022-08-11
Translate deprecations

cobbal:


mstokercricut (approved):
Looks good. This will be important for proper maintenance of our shared libs.

PR #82, opened 2022-11-21
Start of ARM support

bscothern-Cricut:
Now it just needs fat binaries created so they can be linked in and used


cobbal:
Incorporated into #84

PR #85, opened 2022-11-28
Allow use of qualified identifiers

cobbal:


dstoker-cricut (approved):
Looks good to me. Go for it (assuming the coverage failure isn't a problem)

PR #90, opened 2023-01-06
Use fully qualified names in more places

cobbal:
Fixes a kotlin issue in CriGeo-bindings


mstokercricut (approved):
Looks good. Any reason this is only for 4-ary tuples, and not also for 2-ary and 3-ary as well?

cobbal:
> Looks good. Any reason this is only for 4-ary tuples, and not also for 2-ary and 3-ary as well?

Those were already tested. It wasn't a tuple issue that was fixed, it was a list/set/dictionary issue. Tuple just happened to trigger it.

PR #93, opened 2023-01-25
fix names of shared objects in android lib install

cobbal:
FishyJoes was copying apple dylibs into android install locations due to too aggressive refactoring of the build script


steelenelson (approved):
That should work a little better.

PR #96, opened 2023-02-22
Use buttaface's android toolchain in docker

cobbal:


.github/workflows/integration-tests.yaml
```diff
@@ -143,19 +160,83 @@ jobs:
...
+          ./gradlew publishToMavenLocal
+
+      - name: Upload artifacts
+        uses: actions/upload-artifact@v2
+        with:
+          name: maven-local
+          path: ~/.m2/repository
+
+  android-tests:
+    name: "android: integration tests"
```
mstokercricut, line 196: Nit: Should look like the rest: 
`name: Android Integration Tests`

.github/workflows/integration-tests.yaml
```diff
@@ -143,19 +160,83 @@ jobs:
           rm -rf $libdir
           mkdir -p $libdir
 
-          swift build --configuration release --product FishyJoesJavaRuntime
-          cp .build/release/libFishyJoesJavaRuntime.so $libdir
+          ./scripts/compile-kotlin-native-runtime.sh
+
+
+      - name: Build android FishyJoesJavaRuntime
+        uses: addnab/docker-run-action@4f65fabd2431ebc8d299f8e5a018d79a769ae185
```
mstokercricut, line 167: Any concern about security on this one? Looks like it comes from a user, not GitHub...

mstokercricut (approved):
Love the android test app. Approved with nits.

.github/workflows/integration-tests.yaml
cobbal: I looked over the script, and that's why it's pinned to an exact commit

PR #97, opened 2023-03-01
Fix Java Pointers

ejohn-cricut:


cobbal (approved):
Wonderful

Sources/FishyJoesJavaRuntime/Utils.swift
```diff
@@ -101,3 +101,7 @@ public class CStringBag {
         strings = []
     }
 }
+
+public func javaPointer<T>(_ box: Box<T>) -> jvalue {
+    jvalue(j: jlong(Int64(bitPattern: UInt64(UInt(bitPattern: box.retainedOpaque())))))
```
cobbal: On second thought, this function should probably be named something like "retainedJavaPointer" to avoid creating accidental memory leaks.

PR #108, opened 2023-05-12
Update Readme file

ishulumba-ecw:
Implements AB#[Update Readme](https://cricut.visualstudio.com/Cricut/_workitems/edit/334527)
#### Description
Update Readme file


cobbal:
Thanks for the PR! I took some of the information you found, figured out where authentication was failing, verified the auth steps with a clean install, and tried to centralize the changes into a new set of dependency setup instructions in #110 . Hopefully the next person will have an easier time getting started. Thanks again for the feedback.

PR #109, opened 2023-05-16
Bscothern/async

bscothern-Cricut:


Sources/FishyJoesNodeRuntime/NodeReference.swift
```diff
@@ -1,22 +1,27 @@
...
+    public func deallocate(env: NAPI.Env) {
+        guard ref.ptr != nil else {
+            return
+        }
         try? env.deleteReference(ref)
         ref.ptr = nil
     }
+
+    deinit {
+//        precondition(ref.ptr == nil, "Node reference has not been deallocated before going out of scope")
```
bscothern-Cricut: What do we want to do in this case? Does it make sense to dispatch to the main thread and have it deallocate the reference since it has an environment?

Sources/FishyJoesCore/Translators/NodeTranslate.swift
```diff
@@ -121,56 +121,80 @@ struct NodeTranslator: Translator {
...
+                                fragment.output("let convertedTaskResult: NAPI.Value")
+                                fragment.outputBlock("do {", newLineTerminated: false) {
+                                    fragment.output("convertedTaskResult = try \(returnType.converterType.name).toNode(taskResult, env: env)")
+                                }
+                                fragment.outputBlock(" catch {") {
+                                    if method.isMutating {
+                                        fragment.output("try Self.mutateNode(mutatingSelf, this: env.this(), env: env)")
+                                    }
+                                    fragment.output("try env.rejectDeferred(deferred, String.toNode(error.localizedDescription, env: env))")
+                                    fragment.output("throw error")
```
cobbal: This throw feels like it duplicates the error. Can we get rid of it?

Sources/FishyJoesCore/Translators/NodeTranslate.swift
```diff
@@ -121,56 +121,80 @@ struct NodeTranslator: Translator {
...
+                                }
+                                if method.isMutating {
+                                    fragment.output("try Self.mutateNode(mutatingSelf, this: env.this(), env: env)")
```
cobbal: Since this is common with the catch block, maybe it can be de-duplicated and hoisted to the 'do' block

Sources/FishyJoesCore/Translators/NodeTranslate.swift
```diff
@@ -121,56 +121,80 @@ struct NodeTranslator: Translator {
...
+                                }
+                                fragment.outputBlock("try env.resolveDeferred(") {
+                                    fragment.output("deferred,")
+                                    fragment.output("convertedTaskResult")
+                                }
```
cobbal, line 172: This could also be moved to the do block so that we could fall through from the catch block. Your preference

Sources/FishyJoesCore/Translators/NodeTranslate.swift
```diff
@@ -339,7 +359,16 @@ struct NodeTranslator: Translator {
...
+
+            func makeAsyncIfNeeded(_ parameterType: BetterType) -> BetterType {
+                if case let .function(args, returnType, _) = parameterType {
+                    return .function(args, returnType, isAsync: true)
+                } else {
+                    return parameterType
+                }
+            }
+
+            let resolved = context.resolve(type: method.isAsync ? makeAsyncIfNeeded(parameter.type) : parameter.type, generics: exportAnnotation.genericOverrides)
```
cobbal: It seems wrong to silently convert function arguments to async. Are you sure this is still needed?

Sources/FishyJoesNodeRuntime/NAPI.swift
```diff
@@ -7,40 +7,15 @@ public enum NAPI {
...
     public typealias CallbackInfo = swift_napi_callback_info
-    public typealias Deferred = swift_napi_callback_info
+    public typealias Deferred = swift_napi_deferred
 
     public typealias Finalize = napi_finalize
     public typealias CleanupHook = @convention(c) (UnsafeMutableRawPointer?) -> Void
     public typealias Callback = napi_callback
-    public typealias ThreadsafeFunction = napi_threadsafe_function
-    public typealias ThreadsafeReceiveValueFunction = napi_threadsafe_function_call_js
+    public typealias MainThreadCallback = napi_threadsafe_function_call_js
```
cobbal: the main thread callback is a concern of how we're using NAPI, not of NAPI itself, so this typealias doesn't belong here

Sources/FishyJoesNodeRuntime/NodeFunction.swift
```diff
@@ -5,6 +5,86 @@ protocol AnyFunction {
...
+            },
+            continuationBoxPointer
+        )
+        promise = try env.callFunction(promise, catchFunction, [catchCallback])
+    }
+}
+
+private enum PromiseConverter<R>: Converter where R: NodeConverter {
+    static func fromNode(value promise: NAPI.Value, env: NAPI.Env, continuation: CheckedContinuation<R.SwiftType, Error>) throws {
+    }
```
cobbal: Is this intended to be blank?

Sources/FishyJoesNodeRuntime/NodeFunction.swift
```diff
@@ -103,13 +260,42 @@ extension Function0Converter: NodeConverter where R: NodeConverter {
         }
     }
 
-    public static func toNode(_ value: @escaping SwiftType, env: NAPI.Env) throws -> NAPI.Value {
+    public static func toNode(_ value: @escaping SwiftType, env: NAPI.Env) throws 	-> NAPI.Value {
```
cobbal: whitespace

Sources/FishyJoesNodeRuntime/NodeReference.swift
cobbal: Yes, that makes sense now that we have a good way to always get an environment

Sources/FishyJoesNodeRuntime/Utils.swift
```diff
@@ -32,6 +32,26 @@ public func callbackBody(
     }
 }
 
+ public func callbackBody(
```
cobbal: I don't like the name shadowing which can silently introduce an async context. Both this and rethrowToNode should have different names so that the typechecker isn't the only person who knows when an async scope is being introduced

cobbal: Especially since the (node) type is different between the two, as one is a promise.

Sources/FishyJoesNodeRuntime/js/wasm-napi.js
```diff
@@ -1335,22 +1335,22 @@ export class NAPI {
         delete this.references[refIdx];
         return NAPI_OK;
       },
-      napi_create_threadsafe_function: this.wrap((envPtr, func, asyncResourceIdx, asyncResourceNameIdx, maxQueueSize, initialThreadCount, finalizeData, finalizeCallback, context, callJavascriptCallback, resultPtr) => {
+      napi_create_threadsafe_function: this.wrap((envPtr, funcIdx, asyncResourceIdx, asyncResourceNameIdx, maxQueueSize, initialThreadCount, finalizeData, finalizeCallback, callJavascriptMainThreadCallbackContext, callJavascriptMainThreadCallback, resultPtr) => {
```
cobbal: These should follow NAPI for naming, not main thread

cobbal (changes requested):
Looks good

Sources/FishyJoesCore/Translators/NodeTranslate.swift
bscothern-Cricut: Good catch

Sources/FishyJoesCore/Translators/NodeTranslate.swift
bscothern-Cricut: It can't do that because if the resolve throws then we can't call reject on the same `deferred`.

Sources/FishyJoesCore/Translators/NodeTranslate.swift
bscothern-Cricut: Ya not needed anymore.

Sources/FishyJoesNodeRuntime/js/wasm-napi.js
bscothern-Cricut: Fixed.

Sources/FishyJoesCore/Translators/NodeTranslate.swift
bscothern-Cricut: It can't because it needs to be applied with either a reject or resolve of a promise which can't be hoisted up into the do block and fall through.

Sources/FishyJoesNodeRuntime/NodeReference.swift
```diff
@@ -20,7 +20,9 @@ public final class NodeReference: @unchecked Sendable {
     }
 
     deinit {
-//        precondition(ref.ptr == nil, "Node reference has not been deallocated before going out of scope")
+        try? onMainThread { [ref] env in
+            try? env.deleteReference(ref)
```
cobbal: This should guard against nil, probably before jumping threads

cobbal (changes requested):
1 small, but important change needed

Sources/FishyJoesNodeRuntime/NodeReference.swift
bscothern-Cricut: Good catch

cobbal (approved):

PR #113, opened 2023-07-05
Attributed string

mstokercricut:
* Adds translation of AttributedString, AttributedSubstring, AttributeContainer, and a notion of AttributeScope to Kotlin
* Adds translation of URL to Kotlin
* Provides testing for these in TestAPI


Package.swift
```diff
@@ -26,7 +26,7 @@ func wasmIncompatible<T>(_ things: @autoclosure () -> [T]) -> [T] {
 
 let package = Package(
     name: "FishyJoes",
-    platforms: [.macOS(.v11)],
+    platforms: [.macOS(.v12), .iOS(.v15)],
```
cobbal, line 29: Do we support iOS in any way? (Maybe we will need to some day for flutter)

Sources/FishyJoesCore/TranslatedTypes/TranslatedAttributedString.swift
```diff
@@ -0,0 +1,157 @@
...
+    let containedNamedTypes: [TranslatedType] = []
+    let neutralName = "AttributedSubstring.Runs.Run"
+    let nodeName = "AttributedSubstring.Runs.Run"
+    let kotlinPackage: String? = "com.cricut.fishyjoes.runtime"
+    let kotlinName = "AttributedSubstring.Runs.Run"
+    let jniType = JNIType.object("com/cricut/fishyjoes/runtime/AttributedString$Runs$Run")
+    let cSharpType = CSharpClass.CSType.named(package: "Cricut.FishyJoesRuntime", name: "AttributedString.Runs.Run")
+    let cppName = "AttributedString::Runs::Run"
+    let definingModule = Module.runtime
+}
```
cobbal, line 157: Do these need to be all different types? They seem very similar, and could possibly be just different instances of one type to reduce duplication.

cobbal: Similar to, but probably not the same as PrimitiveType

Sources/FishyJoesJavaRuntime/AttributedString/AttributeScope+java.swift
```diff
@@ -0,0 +1,236 @@
...
+            self.link = container.link
+            self.languageIdentifier = container.languageIdentifier
+        }
+    }
+}
+
+extension AttributeContainer {
+    public init(_ foundationAttributes: FoundationAttributes) {
+        self = AttributeContainer()
+        if let link = foundationAttributes.link { self.link = link }
```
cobbal: nit: I kept trying to read this as trailing closure syntax. I think I don't like one-liner if-lets

cobbal (approved):
Approved with nits

Sources/FishyJoesCore/TranslatedTypes/TranslatedAttributedString.swift
mstokercricut: I'll take a look, but I think they do. Each type has pretty different values for each field. I can drop the AttributeScope type, though. The merge policy type could go, too.

Package.swift
mstokercricut: That is one place. I also want to know if iOS lacks support for anything here, even if we don't directly support it, as it informs us where inconsistencies in interface might pop up.

Sources/FishyJoesJavaRuntime/AttributedString/AttributeScope+java.swift
mstokercricut: 👍🏿

Sources/FishyJoesCore/TranslatedTypes/TranslatedAttributedString.swift
mstokercricut: Removed in favor of a boolean.

PR #117, opened 2023-07-19
Fixed JS Exception propigation for async functions

bscothern-Cricut:
Fixes the issue where an async function takes a function as an argument and the provided function throws. Before the error would be wrapped up and lost instead of being properly passed along.


integration-tests/TestAPI-bindings/Sources/Generated/NodeInterface/Functions+node.swift
```diff
@@ -149,6 +149,11 @@ extension Functions: FishyJoesNodeRuntime.NodeConverter {
                                         }
                                         try env.resolveDeferred(deferred, convertedTaskResult)
                                     }
+                                } catch let error as JSException {
+                                    try onMainThread { env in
+                                        let error = try env.createError(NAPI.Value(ptr: nil), String.toNode(error.message, env: env))
```
cobbal, line 154: It's a little sad that we're taking a real JS exception, and converting it to a string before turning that back into a JS exception. Maybe in the future we could pass the original back. Maybe that would be more of a pain than it's worth

cobbal (approved):

integration-tests/TestAPI-bindings/Sources/Generated/NodeInterface/Functions+node.swift
bscothern-Cricut: Ya I couldn't figure out how to wrap it back into a value directly... I'll take another look when I get some time. I'll add a todo on my current working branch.

PR #118, opened 2023-07-19
fix: correct path for copying js extensions and typings

kellyo-cricut:
tested in CriTrace-bindings and successfully pulled in CriRaster extensions

### DIFF
--- a/Sources/FishyJoesExecute/CodeGen.swift
+++ b/Sources/FishyJoesExecute/CodeGen.swift
@@ -303,7 +303,7 @@ extension CodeGen {
 
                     var tsSources = ["Sources/Generated/NodeInterface/\(config.module).d.ts"]
                     for (moduleName, modulePath) in dependencyPaths {
-                        let path = "\(modulePath)/ts/\(moduleName).extensions.d.ts"
+                        let path = "\(modulePath.dropLast("/Sources".count))/ts/\(moduleName).extensions.d.ts"
                         if cmd("test", "-f", path).runBool() {
                             tsSources.append(path)
                         }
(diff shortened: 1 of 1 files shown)


Sources/FishyJoesExecute/CodeGen.swift
```diff
@@ -303,7 +303,7 @@ extension CodeGen {
 
                     var tsSources = ["Sources/Generated/NodeInterface/\(config.module).d.ts"]
                     for (moduleName, modulePath) in dependencyPaths {
-                        let path = "\(modulePath)/ts/\(moduleName).extensions.d.ts"
+                        let path = "\(modulePath.dropLast("/Sources".count))/ts/\(moduleName).extensions.d.ts"
```
cobbal: This seems very fragile. I'd at least feel better if you added `precondition(modulePath.hasSuffix("/Sources"))` to make errors more obvious

cobbal (changes requested):

Sources/FishyJoesExecute/CodeGen.swift
kellyo-cricut: Is that preferable to an `if this ends with "/Sources"` check? [That suffix is added when the dictionary is constructed](https://github.com/cricut/FishyJoes/blob/6785077b65d80dc32771f29742c20c645fee6760/Sources/FishyJoesExecute/CodeGen.swift#L132)

cobbal: Maybe we should add the "Sources" suffix only when it's needed then

cobbal: Or you could use "modulePath/../ts"

cobbal (approved):

PR #123, opened 2023-08-15
C# + Some Node.js Multi-Module Work

mstokercricut:


.github/workflows/c-sharp-runtime.yaml
```diff
@@ -41,13 +41,13 @@ jobs:
...
 
       - name: Upload artifacts
-        uses: actions/upload-artifact@v2
+        uses: actions/upload-artifact@v3
         with:
           name: mac_generated
           path: c-sharp-runtime/runtimes
 
-  build-test-c-sharp-linux:
-    name: "linux: Build and test C# runtime"
```
cobbal, line 50: If you must make all these changes to CI, be sure to do it to the templates in `Sources/FishyJoesExecute/Resources/bindings-template/.github/workflows` as well

.github/workflows/integration-tests.yaml
```diff
@@ -128,59 +143,54 @@ jobs:
           sudo mkdir -p /Library/Developer/Toolchains/swift-wasm-5.7.1-RELEASE.xctoolchain
           sudo tar -xf swift-wasm.tgz -C /Library/Developer/Toolchains/swift-wasm-5.7.1-RELEASE.xctoolchain --strip-components 1
 
-      - name: Test from dotnet
+      - name: Test .NET
```
cobbal: "dotnet" is the name of the executable loading the library. Very analogous to "node"

.github/workflows/integration-tests.yaml
```diff
@@ -128,59 +143,54 @@ jobs:
...
-        with:
-          username: cricut-builduser
-          password:  ${{ secrets.CRICUT_GPR_TOKEN }}
-          registry: ghcr.io
-          image: ghcr.io/cricut/android-swift-runtime:0.1.0
-          options: -v ${{ github.workspace }}:/work -w /work
-          run: /work/scripts/compile-android-runtime.sh
-
-      - name: Test from gradle
+      - name: Test Kotlin
```
cobbal: The "from" parts are very important, because sometimes we test from swift test and sometimes we test from the foreign languages test framework.

.github/workflows/integration-tests.yaml
```diff
@@ -217,7 +227,17 @@ jobs:
...
-      - name: Build Android test app
+      - name: Build Android FishyJoesJavaRuntime
+        uses: addnab/docker-run-action@4f65fabd2431ebc8d299f8e5a018d79a769ae185
+        with:
+          username: cricut-builduser
+          password:  ${{ secrets.CRICUT_GPR_TOKEN }}
+          registry: ghcr.io
+          image: ghcr.io/cricut/android-swift-runtime:0.1.0
+          options: -v ${{ github.workspace }}:/work -w /work
+          run: /work/scripts/compile-android-runtime.sh
```
cobbal: We should be pulling the compiled version from the previous run, not re-building

.github/workflows/integration-tests.yaml
```diff
@@ -231,7 +251,7 @@ jobs:
           cd integration-tests/android-test-app
           ./gradlew assembleAndroidTest
 
-      - name: Test Android app
+      - name: Test Android
```
cobbal: We're testing from the app specifically.

.github/workflows/java-runtime.yaml
```diff
@@ -141,19 +142,25 @@ jobs:
...
-          swift-version: "5.4.2"
+          branch: swift-5.6.2-release
+          tag: 5.6.2-RELEASE
+
+      - name: Build Native FishyJoesJavaRuntime
+        shell: zsh {0}
+        run: |
+          set -euxo pipefail
+
+          CONFIGURATION=debug ./scripts/compile-kotlin-native-runtime.sh
```
cobbal: We should not be shipping debug builds

Sources/FishyJoesCSharpRuntime/AttributedString/AttributeContainer+cSharp.swift
```diff
@@ -0,0 +1,73 @@
...
+    other: AttributeContainer.CType,
+    keepCurrent: Bool.CType,
+    _exn: csOutExn
+) -> VoidConverter.CType {
+    FishyJoesCSharpRuntime.Env.catching(to: _exn) {
+        return try AttributeContainer.mutateCSharp(_cSharpThis) { mutatingSelf in
+            return try VoidConverter.toCSharp(
+                mutatingSelf.merge(
+                    try AttributeContainer.peekCSharp(other),
+                    mergePolicy: !Bool.peekCSharp(keepCurrent) ? .keepNew : .keepCurrent
```
cobbal, line 41: nit: confusing needless negation

Sources/FishyJoesCSharpRuntime/AttributedString/AttributedString.Runs+cSharp.swift
```diff
@@ -0,0 +1,129 @@
...
+        return try Bool.toCSharp(AttributedString.Runs.peekCSharp(lhs) == AttributedString.Runs.peekCSharp(rhs))
+    }
+}
+
+@_cdecl("__cs_AttributedString_Runs_hash")
+public func AttributedString_Runs_cSharpHash(this: csObject, exn: csOutExn) -> Int32.CType {
+    Env.catching(to: exn) {
+        let runHashes = try AttributedString.Runs.peekCSharp(this)
+            .lazy
+            .map { AttributedString("HASH", attributes: $0.attributes).hashValue }
```
cobbal, line 121: Is this ignoring the string component?

Sources/FishyJoesCSharpRuntime/AttributedString/AttributedString.Runs.Run+cSharp.swift
```diff
@@ -0,0 +1,63 @@
...
+    Env.catching(to: exn) {
+        return try Bool.toCSharp(AttributedString.Runs.Run.peekCSharp(lhs) == AttributedString.Runs.Run.peekCSharp(rhs))
+    }
+}
+
+@_cdecl("__cs_AttributedString_Runs_Run_hash")
+public func AttributedString_Runs_Run_cSharpHash(this: csObject, exn: csOutExn) -> Int32.CType {
+    Env.catching(to: exn) {
+        try Int32.toCSharp(
+            Int32(truncatingIfNeeded: AttributedString("HASH", attributes: AttributedString.Runs.Run.peekCSharp(this).attributes).hashValue)
```
cobbal, line 60: and here

Sources/FishyJoesCSharpRuntime/Env.swift
```diff
@@ -60,8 +66,9 @@ public enum Env {
 
     public static func unwrap<R>(_ value: R?, file: StaticString = #file, line: UInt = #line) throws -> R {
         guard let value = value else {
-            var message = ["\(file):\(line): Unexpected null"]
-            Thread.callStackSymbols.forEach { message.append($0) }
+            let message = ["\(file):\(line): Unexpected null"]
+            print(message)
```
cobbal, line 70: Probably shouldn't be an array anymore if only 1 line

scripts/compile-c-sharp-runtime.sh
```diff
@@ -19,8 +19,22 @@ else
     BIN_DIR=$(swift build --configuration $CONFIGURATION --show-bin-path)
 fi
 
-mkdir -p c-sharp-runtime/runtimes/{osx,win,linux}/native
+function install-lib () {
```
cobbal: style: Prefer function variable passing (referring to $1 and $2 inside the function) to pass-by-global-variable

scripts/compile-c-sharp-runtime.sh
```diff
@@ -19,8 +19,22 @@ else
...
+LIB_DIR=c-sharp-runtime/runtimes/win/native
+install-lib
+
+LIB_NAME="libFishyJoesCSharpRuntime.dylib"
+LIB_DIR=c-sharp-runtime/runtimes/osx/native
+install-lib
+
+LIB_NAME="libFishyJoesCSharpRuntime.so"
+LIB_DIR=c-sharp-runtime/runtimes/linux/native
+install-lib
```
cobbal: You've changed behavior here. Previously, at least 1 install had to succeed for the script to succeed. Now if all fail, the script still succeeds.

cobbal (changes requested):

scripts/compile-c-sharp-runtime.sh
cobbal: (I suggest keeping the `||`s, but using them between the function invocations)

mstokercricut: 👍🏿

Sources/FishyJoesCSharpRuntime/AttributedString/AttributedString.Runs+cSharp.swift
mstokercricut: Not sure. You mentioned that all empty attributed strings were equal, so I put in a non-empty string. I'll check the actual hash values to ensure they have a good spread.

cobbal: For the run we should have the actual string contents as well as the attributes, shouldn't we? Or am I misunderstanding runs?

cobbal (approved):

PR #125, opened 2023-08-23
Modules almost everywhere

cobbal:
Put modules into more typenames more places


Sources/FishyJoesCore/BetterType.swift
```diff
@@ -5,8 +5,9 @@ import SourceryRuntime
 // An Optional<Int> isn't a Int that happens to be optional, it's an Optional that may contain an Int!
 indirect enum BetterType: Codable, Hashable {
     struct Name: Codable, Hashable {
-        let name: String
+        var module: String?
         var namespace: [String]
+        let name: String
```
mstokercricut, line 10: Is `name` a good identifier for this field anymore? I'm thinking of URLs, which have host, tld, etc. Does a `BetterType.Name` have a similar breakdown? E.g. module, namespace, actual? Or module, namespace, identifier? Maybe module, namespace, concrete?

mstokercricut (approved):
Approved with nit

Sources/FishyJoesCore/BetterType.swift
cobbal: "concrete" is too overloaded already with generics. If anything, we should rename the type "QualifiedName", but both "Name" and "name" were already in use here, they're just re-ordered.

PR #127, opened 2023-08-24
Wasm Support for Runtime

mstokercricut:
* Made changes necessary to reference the node runtime from Wasm
* Documented CodeGen.swift

### DIFF
--- a/Sources/FishyJoesExecute/CodeGen.swift
+++ b/Sources/FishyJoesExecute/CodeGen.swift
@@ -114,8 +114,8 @@ extension CodeGen {
     }
 
     public mutating func run() throws {
+        // Parse swift package information from the Package.swift file for the bindings module
         let packageJSON = try cmd("swift", "package", "dump-package").runData()
-
         let packageInfo: SwiftPackage
         do {
             packageInfo = try JSONDecoder().decode(SwiftPackage.self, from: packageJSON)
@@ -123,6 +123,7 @@ extension CodeGen {
             fatalError("Couldn't parse swift package: \(error)")
         }
 
+        // Locate dependency bindings modules required by this bindings module
         var dependencyPaths: [String: String] = [config.module: "."]
         for moduleName in config.requiredModules {
             let bindingModule = "\(moduleName)-bindings"
@@ -132,30 +133,14 @@ extension CodeGen {
             let dependencyPath = (dependencyURL.scheme == nil ? dependencyURL.path : ".build/checkouts/\(bindingModule)") + "/Sources"
             dependencyPaths[moduleName] = dependencyPath
         }
-
         let localPathsNeeded = packageInfo.dependencyMap.compactMap {
             let url = $0.value
             return url.scheme == nil ? url.path : nil
         }
 
-        let makeDockerContext = useDocker ? {
-            let context = DockerContext(withAvailablePaths: localPathsNeeded)
-            if let context = context {
-                printAndFlush("found docker binary: \(context.hostDockerBinary)")
-            } else {
-                printAndFlush("not using docker")
-            }
-            return context
-        } : { nil }
-
-        let configuration = BuildConfiguration(
-            debug: debug,
-            fat: fat,
-            codeCoverage: codeCoveragePath != nil,
-            baseDockerContext: Lazy(makeDockerContext())
-        )
-
+        // MARK: - Generate Step
         if buildStep.contains(.generate) {
+            // Locate sources to translate
             let translateeSources: String
             if let translateeDependency = packageInfo.dependencyMap[config.module.lowercased()] {
                 translateeSources = (translateeDependency.scheme == nil ? translateeDependency.path : ".build/checkouts/\(config.module)") + "/Sources"
@@ -168,11 +153,12 @@ extension CodeGen {
                 fatalError("Couldn't locate FishyJoes in Package.swift")
             }
 
+            // Locate dependency module configuration files
             let fishyJoesModuleFiles: [String] = dependencyPaths.compactMap {
                 $0.key == config.module ? nil : "\($0.value)/Generated/\($0.key).fishyjoesmodule"
             }
 
-            // MARK: Generate code
+            // Prepare destination directories for the Swift-side interface files for the translated foreign languages
             try cmd("rm", "-rf", "Sources/Generated", "kotlin/src/generated", "DebugGenerated", "cpp").run()
             try cmd("mkdir", "-p",
                     "Sources/Generated/CSharpInterface",
@@ -188,7 +174,11 @@ extension CodeGen {
                 "Sources/Generated/JavaInterface/EmptyPlaceholder.swift",
                 "Sources/Generated/CPPInterface/EmptyPlaceholder.swift"
             ).run()
+
+            // Build the Sourcery tool itself
             try cmd("swift", "build", "--product", "sourcery").run()
+
+            // Build the FishyJoes internal execution helper product
             try cmd("swift", arguments: ["build"] + (codeCoveragePath == nil ? [] : Platform.coverageFlags) + ["--product", "🐟☕️"]).run()
 
             // Trampoline into fishy-joes-execution-helper via Sourcery
@@ -200,7 +190,8 @@ extension CodeGen {
                 sourceryEnv["LLVM_PROFILE_FILE"] = "\(codeCoveragePath)/fishy-joes-execution-helper-\(UUID()).profraw"
             }
 
-            // Provide access to stderr for the fishy-joes core. This is necessary because sourcery intercepts stderr and fails if it is used at all.
+            // Provide access to stderr for the fishy-joes core
+            // This is necessary because sourcery intercepts stderr and fails if it is used at all
             // Create a temporary named pipe
             // https://unix.stackexchange.com/a/29918/5471
             let errorFifoPath = try cmd("mktemp", "-u").runString()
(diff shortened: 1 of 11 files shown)


Sources/FishyJoesExecute/CodeGen.swift
```diff
@@ -132,30 +133,14 @@ extension CodeGen {
...
-        let configuration = BuildConfiguration(
-            debug: debug,
-            fat: fat,
-            codeCoverage: codeCoveragePath != nil,
-            baseDockerContext: Lazy(makeDockerContext())
-        )
-
+        // MARK: - Generate Step
         if buildStep.contains(.generate) {
+            // Locate sources to translate
```
cobbal, line 146: In particular, this is locating dependency sources. e.g. CriGeo (not -bindings) and FishyJoes

Sources/FishyJoesExecute/CodeGen.swift
```diff
@@ -168,11 +153,12 @@ extension CodeGen {
                 fatalError("Couldn't locate FishyJoes in Package.swift")
             }
 
+            // Locate dependency module configuration files
             let fishyJoesModuleFiles: [String] = dependencyPaths.compactMap {
                 $0.key == config.module ? nil : "\($0.value)/Generated/\($0.key).fishyjoesmodule"
             }
 
-            // MARK: Generate code
+            // Prepare destination directories for the Swift-side interface files for the translated foreign languages
```
cobbal: This is cleaning/creating any directories that will be referenced by the sourcery-side generator, not just swift

Sources/FishyJoesExecute/CodeGen.swift
```diff
@@ -228,7 +220,25 @@ extension CodeGen {
...
+            let makeDockerContext = useDocker ? {
+                let context = DockerContext(withAvailablePaths: localPathsNeeded)
+                if let context = context {
+                    printAndFlush("found docker binary: \(context.hostDockerBinary)")
+                } else {
+                    printAndFlush("not using docker")
+                }
+                return context
+            } : { nil }
+            let configuration = BuildConfiguration(
```
cobbal, line 235: I'm a little surprised this move was possible. I didn't realize build was the only step to rely on the configuration.

Sources/FishyJoesExecute/CodeGen.swift
```diff
@@ -271,24 +281,25 @@ extension CodeGen {
...
-                try cmd("rm", "-rf", platform.outputDir(config)).run()
-                try cmd("mkdir", "-p", platform.outputDir(config)).run()
-            }
-
+            // Install libraries to platform-specific output directories
             for platform in platforms {
+                // Prepare output directory
                 let outputDir = platform.outputDir(config)
+                try cmd("rm", "-rf", outputDir).run()
+                try cmd("mkdir", "-p", outputDir).run()
```
cobbal: It's a bug to merge these 2 loops, since outputDir may be the same between 2 platforms.

Sources/FishyJoesExecute/CodeGen.swift
```diff
@@ -271,24 +281,25 @@ extension CodeGen {
...
                     let src = "\(try platform.buildDir(configuration))/lib\(name).\(platform.dylibExt)"
                     let installName = installName ?? "lib\(name).\(platform.dylibExt)"
                     let dest = "\(outputDir)/\(installName)"
                     try cmd("cp", src, dest).run()
                 }
 
+                // Perform library installation and platform-specific customization
                 switch platform {
                 case .wasm:
+                    // Configure wasm-opt to optimize Wasm bundles
```
cobbal: What this block really does is installs the wasm blob into the output. It either uses `wasm-opt` or `cp` depending on settings and installed software.

Sources/FishyJoesExecute/CodeGen.swift
```diff
@@ -301,7 +312,17 @@ extension CodeGen {
                     }
                     try cmd("cp", "\(platform.buildDir(configuration))/FishyJoes_FishyJoesNodeRuntime.resources/js/wasm-napi.js", outputDir).run()
 
-                    var tsSources = ["Sources/Generated/NodeInterface/\(config.module).d.ts"]
+                    // Find the path to the runtime
+                    let runtimeDirectorySearchLocations = [".build/checkouts/FishyJoes/node-runtime", "../../node-runtime"]
```
cobbal: This feels very fragile and is missing at least 1 case. Also, I don't like seeing any testapi-specific config outside of the `integration-tests` directory.

cobbal: Probably grab the result of `packageInfo.dependencyMap["fishyjoes"]` from the much earlier guard statement like we do for other dependencies

Sources/FishyJoesExecute/CodeGen.swift
```diff
@@ -310,19 +331,19 @@ extension CodeGen {
                     }
                     try cmd("cat", arguments: tsSources).output(overwritingFile: "\(outputDir)/\(config.module).d.ts").run()
 
-                    // replace each instance of __MODULE_NAME__ with the name
-                    // replace each line containing __MODULE_DEPENDENCY__ with one line for each dependency
+                    // Create the required Javascript files for loading the module's Wasm bundle from a web browser
```
cobbal: This is both for browser and non-browser

Sources/FishyJoesExecute/CodeGen.swift
```diff
@@ -342,10 +362,12 @@ extension CodeGen {
                         outPath: "\(outputDir)/\(config.module).browser.js"
                     )
 
+                    // Configure loading of Javascript extensions when the Wasm bundle is loaded, if provided by dependency modules
```
cobbal: This doesn't configure the extensions, it just installs them

Sources/FishyJoesExecute/CodeGen.swift
```diff
@@ -410,74 +438,90 @@ extension CodeGen {
...
+            for platform in platforms {
+                switch platform {
+                case .wasm, .node:
+                    break
+                case .kotlinSystem:
+                    try FileManager.default.withCurrentDirectoryPath("kotlin") {
+                        try cmd("./gradlew", "build", "-Dskip.tests").run()
+                    }
+                case .kotlinAndroid:
+                    // TODO: Compile using Android Studio
```
cobbal: This comment is wrong. We compile with gradle exactly as above. Avoiding duplicate compilation is why this was an "if contains". Since kotlinSystem is always present when kotlinAndroids are, the check was simplified to just check for kotlinSystem

Sources/FishyJoesExecute/CodeGen.swift
```diff
@@ -410,74 +438,90 @@ extension CodeGen {
...
-            case .kotlinSystem, .kotlinAndroid, .cpp, .cSharp:
-                break
             }
         }
 
-        // MARK: test that things run properly
+        // MARK: - Test Step
         if buildStep.contains(.test) {
             for platform in platforms {
+                // Gather environment information to pass on for code-coverage purposes
```
cobbal: "environment variables" is more precise and technical than "information"

Sources/FishyJoesExecute/CodeGen.swift
```diff
@@ -526,19 +572,30 @@ extension CodeGen {
...
             let version = version ?? "0.0.1-unknown"
             for platform in platforms {
-                if platform.isTs {
+                switch platform {
+                case .wasm, .node:
+                    // Pack using npm
                     try cmd("npm", "pack", "./\(platform.outputDir(config))").run()
-                } else if platform == .cSharp {
+                case .kotlinSystem, .kotlinAndroid:
+                    // TODO: Pack using maven
```
cobbal: nothing to do here

Sources/FishyJoesExecutionHelper/main.swift
```diff
@@ -12,7 +12,16 @@ if let dumpPath = ProcessInfo.processInfo.environment["DUMP_SOURCERY_DATA"] {
 }
 
 let inputPath = (CommandLine.arguments[2] as NSString).expandingTildeInPath as String
-guard let context = NSKeyedUnarchiver.unarchiveObject(withFile: inputPath) as? TemplateContext else {
+protocol UndeprecatedNSKeyedUnarchiver {
+    static func unarchiveObject(withFile file: String) -> Any?
+}
+enum DeprecatedNSKeyedUnarchiver: UndeprecatedNSKeyedUnarchiver {
```
cobbal, line 18: Nice workaround. The enum is unneccesary though, you can directly do
```
extension NSKeyedUnarchiver: UndeprecatedNSKeyedUnarchiver {}
```

cobbal (changes requested):

Sources/FishyJoesExecute/CodeGen.swift
mstokercricut: Given its name `BuildConfiguration` I was pleased to see it. 😁

Sources/FishyJoesExecutionHelper/main.swift
mstokercricut: Not sure what's going on, but using an extension causes Sourcery to puke: 

Generating code...
error: .build/debug/FishyJoes_FishyJoesExecutionHelper.bundle/FishyJoes.swifttemplate: Swift/ErrorType.swift:200: Fatal error: Error raised at top level: SwiftTemplate.ProcessError(description: "xcrun [\"--executing-from-within-sourcery\", \"/var/folders/wc/svg0n55n0rv5fj3y71npvjjm0000gp/T/SwiftTemplate/1.9.2/context.bin\"] exited with code 4")

cobbal: Works on my machine :shrug:

cobbal (approved):

PR #130, opened 2023-09-05
Fix wrong module being attached to types defined in extensions

cobbal:
PR #125 broke CriText because it defines new types under the swift UnicodeScalar namespace.

Fix by trying to resolve the root of the namespace, and using its module instead of the module where the type is defined.


bscothern-Cricut (commented):
LGTM

PR #132, opened 2023-09-12
🎯 Dart

cobbal:


.github/workflows/c-sharp-runtime.yaml
```diff
@@ -33,12 +33,12 @@ jobs:
         with:
           credentials: https://cricut-builduser:${{ secrets.CRICUT_GPR_TOKEN }}@github.com/
 
-      - name: Build FishyJoesCSharpRuntime
+      - name: Build FishyJoesIotaRuntime
```
mstokercricut, line 36: Since CSharp / Dart never got shared, it would be less confusing for future readers if a project-wide search / replace for `Iota` to `Dart` was applied.

Sources/FishyJoesCore/FishyJoesContext.swift
```diff
@@ -14,19 +14,26 @@ public class FishyJoesContext {
...
+    private(set) var kotlinClasses: [KotlinClass] = []
+    private(set) var cSharpClasses: [CSharpClass] = []
+    private(set) var dartClasses: [DartClass] = []
 
     let nodeTranslator = NodeTranslator()
     let kotlinTranslator = KotlinTranslator()
-    let cSharpTranslator = CSharpTranslator()
     let neutralTranslator = NeutralTranslate()
 
+    let iotaTranslator = IotaTranslator()
```
mstokercricut, line 25: Oh, are there still some common components? Are they significant enough to warrant the confusion another segmentation brings?

Sources/FishyJoesCore/FishyJoesContext.swift
```diff
@@ -311,22 +347,22 @@ public class FishyJoesContext {
...
-            "UInt8": (c: "uint8_t", ts: "number", jni: JNIType.byte, cSharp: "byte"),
-            "UInt16": (c: "uint16_t", ts: "number", jni: JNIType.short, cSharp: "ushort"),
-            "UInt32": (c: "uint32_t", ts: "number", jni: JNIType.int, cSharp: "uint"),
-            "UInt64": (c: "uint64_t", ts: "bigint", jni: JNIType.long, cSharp: "ulong"),
-            "UInt": (c: "uint64_t", ts: "number", jni: JNIType.long, cSharp: "nuint"),
+            "UInt8": (c: "uint8_t", ts: "number", jni: JNIType.byte, cSharp: "byte", dart: "int", dartFFI: "Uint8"),
+            "UInt16": (c: "uint16_t", ts: "number", jni: JNIType.short, cSharp: "ushort", dart: "int", dartFFI: "Uint16"),
+            "UInt32": (c: "uint32_t", ts: "number", jni: JNIType.int, cSharp: "uint", dart: "int", dartFFI: "Uint32"),
+            "UInt64": (c: "uint64_t", ts: "bigint", jni: JNIType.long, cSharp: "ulong", dart: "int", dartFFI: "Uint64"),
+            "UInt": (c: "uint64_t", ts: "number", jni: JNIType.long, cSharp: "nuint", dart: "int", dartFFI: "UnsignedInt" /* Seriously, dart? */),
```
mstokercricut, line 365: 🤦

Sources/FishyJoesCore/Unparse/DartClass.swift
```diff
@@ -0,0 +1,663 @@
...
+            outputNativeMethodDeclarations(to: fragment)
+        }
+    }
+}
+
+extension DartClass {
+    private static var forbiddenVarNames: Set<String> = [
+    ]
+
+    static func deforbidify(_ name: String) -> String {
```
mstokercricut, line 652: This... is a verb that makes a surprising amount of sense by itself. Nice!

Sources/FishyJoesCore/Unparse/DartClass.swift
```diff
@@ -0,0 +1,663 @@
...
+extension DartClass.DartType: CustomStringConvertible {
+    static var object: Self {
+        .named(package: nil, name: "object")
+    }
+
+    var description: String {
+        "FIXME: You should not use this, you should use one of the representations below. \(name())"
+    }
+
+    // static func pretendDartHasNamespaces(name: String) -> String {
```
mstokercricut: Still need this?

Sources/FishyJoesCore/Utilities.swift
```diff
@@ -33,6 +34,10 @@ func upperCaseFirst<S: StringProtocol>(_ camel: S) -> String {
     (camel.first?.uppercased() ?? "") + camel.dropFirst()
 }
 
+func fakeDartNamespace<S: StringProtocol>(_ name: S) -> String {
```
mstokercricut: Nit: Seems like this should belong in the dart-specific utilities, or be renamed to something more generic like `collapsedName(forNamespace:)`

Sources/FishyJoesExecute/Platform.swift
```diff
@@ -51,6 +52,8 @@ enum Platform: CustomStringConvertible, Hashable {
             return "kotlinAndroid(\(arch.rawValue))"
         case .cSharp:
             return "cSharp"
+        case .dartSystem:
```
mstokercricut: Is there a non-system variant (like kotlin android)? If not, could this just be `.dart`?

dart-runtime/bin/setup.dart
```diff
@@ -0,0 +1,389 @@
+import "dart:io" as io;
+import "dart:convert" as convert;
+import 'package:path/path.dart' as path;
+import 'package:yaml/yaml.dart' as yaml;
+
+class Download {
```
mstokercricut, line 6: Ugg. Is the Dart standard 2 spaces?

dart-runtime/lib/utilities.dart
```diff
@@ -0,0 +1,166 @@
...
+
+/// <summary> A reference that becomes owned by the callee </summary>
+typedef ConsumedRef = ffi.Pointer<_ConsumedRef>;
+
+typedef OutCreatedRef = ffi.Pointer<CreatedRef>;
+
+class _Env extends ffi.Opaque {}
+typedef Env = ffi.Pointer<_Env>;
+
+class _GCWhisperer {
```
mstokercricut, line 28: GCs are a skiddish lot. Got to approach them calmly and quietly.

integration-tests/TestAPI-bindings/Sources/Generated/CSharpInterface/TestAPI.Collections+cSharp-type.swift
```diff
@@ -1,20 +1,30 @@
+<<<<<<<< HEAD:integration-tests/TestAPI-bindings/Sources/Generated/IotaInterface/Collections+iota-type.swift
```
mstokercricut: Does this still build with these here?!?

integration-tests/TestAPI-bindings/Sources/Generated/IotaInterface/Swift.UnicodeScalar.PuttingTypesIntoQuestionablePlaces+iota-type.swift
```diff
@@ -0,0 +1,57 @@
+// Generated using Sourcery 2.0.2 — https://github.com/krzysztofzablocki/Sourcery
```
mstokercricut, line 1: Ya... UnicodeScalar is not a great umbrella. Thoughts on a better place?

mstokercricut (changes requested):
This is very good stuff. Fits surprisingly well into the existing systems. I wanted to be sure we want to keep Iota instead of having some duplicated code between C# and Dart. It seems like this might end out being a headache where we change something for Dart and it affects C#'s operation, etc. The isolation between languages may be worth the duplication as the codebase and requirements grow. I'm not set on separation, but I'm leaning that way with the area of overlap being where it is.

Sources/FishyJoesCore/FishyJoesContext.swift
cobbal: I re-unified the 2, after figuring everything out. It essentially boiled down to adding a "void * context" style parameter to many methods. It really saves a lot of nearly-but-not-quite duplication that was very annoying to develop.

dart-runtime/bin/setup.dart
cobbal: "Indent block and collection bodies two spaces." "Indent continued lines with at least four spaces."

Really the canonical formatting is whatever "dart fmt" does to your code. It's really ugly though, and I stopped using it...

cobbal: https://github.com/dart-lang/dart_style/wiki/Formatting-Rules

dart-runtime/lib/utilities.dart
cobbal: "GCCoordinator" would probably be the more serious name, but I don't think it's particularly more informative than whisperer

integration-tests/TestAPI-bindings/Sources/Generated/IotaInterface/Swift.UnicodeScalar.PuttingTypesIntoQuestionablePlaces+iota-type.swift
cobbal: Not really, but that's something that's been in CriText for a while now.

mstokercricut (approved):

PR #140, opened 2023-10-16
Attributed string reviewed

mstokercricut:


.github/workflows/integration-tests.yaml
```diff
@@ -15,64 +15,89 @@ concurrency:
...
+      - name: Build FishyJoesNodeRuntime
         shell: zsh {0}
         run: |
           set -euxo pipefail
-          cd integration-tests/TestAPI-bindings
 
-          npm install
+          ./scripts/compile-node-runtime.sh
+
+      #TODO: Wasm
```
cobbal: This shouldn't be a TODO. We don't build wasm on mac anymore

.github/workflows/integration-tests.yaml
```diff
@@ -123,22 +149,32 @@ jobs:
...
+      - name: Test TypeScript from npm via FishyJoes
+        shell: zsh {0}
+        run: |
+          set -euxo pipefail
+
           cd integration-tests/TestAPI-bindings
+          #swift run -- fishy-joes --nodejs --wasm build test
```
cobbal: This should be tested

.github/workflows/integration-tests.yaml
```diff
@@ -174,51 +208,55 @@ jobs:
...
+      - name: Test kotlin from gradle via fishy-joes
         shell: zsh {0}
         run: |
           set -euxo pipefail
 
-          (cd kotlin-runtime && ./gradlew publishToMavenLocal)
-
           cd integration-tests/TestAPI-bindings
           swift run -- fishy-joes --kotlin build test
+          (cd kotlin && ./gradlew publishToMavenLocal)
```
cobbal: This doesn't need to be published

.github/workflows/iota-runtime.yaml
```diff
@@ -41,13 +41,13 @@ jobs:
...
           name: mac_generated
           path: .build/apple/release/libFishyJoesIotaRuntime.dylib
 
-  build-c-sharp-linux:
-    name: "linux: Build iota runtime"
+  build-iota-ubuntu:
+    name: "Ubuntu: Build Iota Runtime"
     runs-on: ubuntu-latest
     container:
       image: swift
```
cobbal, line 54: I think this got updated to use the android image in the kotlin workflow to ensure our libswiftwhatever.so ends up with the same version. We should probably do the same here.

.github/workflows/iota-runtime.yaml
```diff
@@ -134,47 +135,47 @@ jobs:
           branch: swift-5.6.2-release
           tag: 5.6.2-RELEASE
 
-      - name: Compile swift
+      - name: Build FishyJoesIotaRuntime
+        shell: pwsh
         run: |
+          # Should perform the same actions as scripts/compile-c-sharp-runtime.sh, but must be done in powershell for swift-build to work
```
cobbal: c-sharp -> iota

.github/workflows/node-runtime.yaml
```diff
@@ -0,0 +1,285 @@
...
+
+      - name: Copy swift runtime DLLs
+        shell: bash
+        run: |
+          set -euxo pipefail
+
+          cp -R /c/Library/icu-*/usr/bin/*.dll $LIB_DIR
+          cp -R /c/Library/Swift-development/bin/*.dll $LIB_DIR
+
+      # TODO: Unit test node runtime instead of relying on integration tests
```
cobbal: Again, not needed. This TODO seems to be copied from CriGeoCLI

.github/workflows/node-runtime.yaml
```diff
@@ -0,0 +1,285 @@
...
+        uses: actions/download-artifact@v3
+
+      - name: Install Artifacts
+        shell: zsh {0}
+        run: |
+          set -euxo pipefail
+
+          ls *
+
+          rm -rf node-runtime
```
cobbal: This contains sources. Ideally only the generated binaries would be copied over from previous steps

.github/workflows/node-runtime.yaml
```diff
@@ -0,0 +1,285 @@
...
+          echo "Detected Version is '$VERSION'"
+          NPM_PACKAGE_VERSION=`if [[ "$VERSION" =~ $VERSION_REGEX ]]; then echo "$VERSION"; else echo "0.0.1"; fi`
+          echo "Detected NPM Package Compatible Version is '$NPM_PACKAGE_VERSION'"
+          echo VERSION=$VERSION >> $GITHUB_ENV
+          echo NPM_PACKAGE_VERSION=$NPM_PACKAGE_VERSION >> $GITHUB_ENV
+
+      - name: Setup Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: '12.x'
```
cobbal: We're using 16 in most places

.github/workflows/node-runtime.yaml
```diff
@@ -0,0 +1,285 @@
...
+#          set -euxo pipefail
+#
+#          cd node-runtime
+#          npm test
+
+      - name: Upload library
+        uses: actions/upload-artifact@v3
+        with:
+          name: windows_generated
+          path: node-runtime
```
cobbal: Only the necessary things should be uploaded

.github/workflows/node-runtime.yaml
```diff
@@ -0,0 +1,285 @@
...
+        run: |
+          set -euxo pipefail
+
+          VERSION=${GITHUB_REF#refs/*/}
+          VERSION_REGEX="[0-9]+\.[0-9]+\.[0-9]+(-[a-zA-Z0-9]+)*"
+          if [[ ! $VERSION =~ $VERSION_REGEX ]]; then
+              VERSION="0.0.1-$(git rev-parse --short HEAD)"
+          fi
+          echo "Detected Version is '$VERSION'"
+          NPM_PACKAGE_VERSION=`if [[ "$VERSION" =~ $VERSION_REGEX ]]; then echo "$VERSION"; else echo "0.0.1"; fi`
```
cobbal: style: turn this into an if statement and just set the variable in both branches

.github/workflows/node-runtime.yaml
```diff
@@ -0,0 +1,285 @@
...
+          registry-url: 'https://npm.pkg.github.com/'
+          scope: '@cricut'
+
+      - name: NPM Pack macOS
+        shell: zsh {0}
+        run: |
+          set -euxo pipefail
+
+          LIB_DIR=node-runtime/fishyjoes-runtime-native-macos
+          sed "s:0.0.1:${{ env.NPM_PACKAGE_VERSION }}:g" ${LIB_DIR}/package.json > ${LIB_DIR}/package.json.upd
```
cobbal: sed is the wrong tool for this job. Either "0.0.1" needs to be replaced with a more unique sentinel, or (preferably) `jq` should be used to set the variable in a json-aware manner

.github/workflows/node-runtime.yaml
```diff
@@ -0,0 +1,285 @@
...
+          mv ${LIB_DIR}/package.json.upd ${LIB_DIR}/package.json
+          cd ${LIB_DIR}
+          ls -la
+          npm pack
+          echo MAC_LIB_DIR=${LIB_DIR} >> $GITHUB_ENV
```
cobbal: setting env vars from inside a script should only be done in cases where we don't know what the variable will be. Instead, set it in the "env" section and just refer to $MAC_LIB_DIR instead of $LIB_DIR

.github/workflows/node-runtime.yaml
```diff
@@ -0,0 +1,285 @@
...
+#          cd node-runtime
+#          npm test
+
+      - name: Upload library
+        uses: actions/upload-artifact@v3
+        with:
+          name: windows_generated
+          path: node-runtime
+
+  node-publish:
```
cobbal: There's not really any need to have this job. Each package is isolated, and doesn't need the others. Instead, packing and publishing can happen in the same job as building

Sources/FishyJoesCore/FishyJoesContext.swift
```diff
@@ -76,7 +76,7 @@ public class FishyJoesContext {
     }
 
     func swiftFragment(_ name: String, additionalImports: [String] = []) -> SourceFragment {
-        var headerLines = (module.dependencies + [module] + additionalImports).map { "import \($0)" }
+        var headerLines = (module.dependencies + [module.description] + additionalImports).map { "import \($0)" }
```
cobbal: Prefer "name" to "description"

Sources/FishyJoesCore/FishyJoesContext.swift
```diff
@@ -385,10 +392,32 @@ public class FishyJoesContext {
                 } else if let typeOverride = generics[name.name] {
                     dontCache = true
                     return try recur(typeOverride)
-                } else if name.name == "String" {
+                } else if name.globalName == "String" {
```
cobbal: Why does this even work? globalName should be Swift.String

Sources/FishyJoesCore/Translators/NodeTranslator.swift
```diff
@@ -227,9 +227,10 @@ struct NodeTranslator: Translator {
         nodeTypeListFragment.output("@_cdecl(\"napi_register_module_v1\")")
         nodeTypeListFragment.outputBlock("public func napi_register_module_v1(env: napi_env, exports: napi_value) -> napi_value? {") {
             nodeTypeListFragment.output("let env = NAPI.Env(ptr: env)")
-            nodeTypeListFragment.output("let exports = NAPI.Value(ptr: exports)")
+            nodeTypeListFragment.output("var exports = NAPI.Value(ptr: exports)")
             nodeTypeListFragment.outputBlock("return FishyJoesNodeRuntime.rethrowToNode(env: env) {") {
-                nodeTypeListFragment.output("try registerModule\(context.module)(env: env, exports: exports)")
+                nodeTypeListFragment.output("exports = try registerModule\(context.module)(env: env, exports: exports)")
+                nodeTypeListFragment.output("return exports")
```
cobbal: These changes do nothing but introduce unused mutation

Sources/FishyJoesExecute/CodeGen.swift
```diff
@@ -111,68 +111,73 @@ extension CodeGen {
...
             platforms.append(.dart)
         }
+
+        // When no platforms are provided, use as default the platforms that can execute natively on the build machine
+        if platforms.isEmpty {
+            platforms.append(.node)
+            platforms.append(.kotlinSystem)
+            platforms.append(.cSharp)
+            platforms.append(.dart)
+        }
```
cobbal, line 117: These are bad defaults. If anything we should do less when no parameters are specified

Sources/FishyJoesExecute/CodeGen.swift
```diff
@@ -111,68 +111,73 @@ extension CodeGen {
...
         let packageInfo: SwiftPackage
         do {
             packageInfo = try JSONDecoder().decode(SwiftPackage.self, from: packageJSON)
         } catch let error {
             fatalError("Couldn't parse swift package: \(error)")
         }
+        func dependencyForModule(_ moduleIdentifier: String) -> URL? {
+            return packageInfo.dependencyMap[moduleIdentifier.lowercased()]
+        }
+        func dependencyIsFileRelative(_ dependencyURL: URL) -> Bool {
```
cobbal: "file relative" is the wrong word for this, since it may or may not be relative

Sources/FishyJoesExecute/CodeGen.swift
```diff
@@ -111,68 +111,73 @@ extension CodeGen {
...
+            return dependencyURL.scheme == nil
+        }
+        func localPath(toModule moduleIdentifier: String) -> String? {
+            return dependencyForModule(moduleIdentifier).map { localPath(toDependency: $0) } ?? nil
+        }
+        func localPath(toDependency dependencyURL: URL) -> String? {
```
cobbal: Too many functions. Dependency should be a type at this point

Sources/FishyJoesExecute/CodeGen.swift
```diff
@@ -437,69 +512,112 @@ extension CodeGen {
...
-                        .prettyPrinted,
-                        .withoutEscapingSlashes
-                    ]
+            // Generate files whose creation requires use of template files
+            for platform in platforms {
+                switch platform {
+                case .wasm, .node:
+                    // Generate package.json from template
+                    let packageVersion = version ?? "0.0.1" // If no version is provided, use a dummy version to build the package
+                    let runtimeVersion = /*fishyJoesDependency.version ?? */"file:\(fishyJoesLocalPath)/node-runtime/fishyjoes-runtime-\(platform.executionEnvironment)" // If fishy-joes is file-local, use a file-local runtime too
```
cobbal: This feels incomplete

Sources/FishyJoesExecute/Platform.swift
```diff
@@ -248,6 +248,13 @@ enum Platform: CustomStringConvertible, Hashable {
         }
     }
 
+    var executionEnvironment: String {
+        return platform
+            .replacingOccurrences(of: "node-native-", with: "native-")
+            .replacingOccurrences(of: "jni-", with: "native-")
```
cobbal: I don't understand at all what this variable means outside of node, so I would be happier if it didn't return a value outside of node.

Sources/FishyJoesExecute/SwiftPackage.swift
```diff
@@ -89,3 +95,92 @@ extension SwiftPackage {
         return target.path ?? "./Sources/\(targetName)"
     }
 }
+
+/*
```
cobbal: Is this a valuable comment?

Sources/FishyJoesNodeRuntime/Utils.swift
```diff
@@ -9,6 +9,14 @@ public struct JSException: Error {
     }
 }
 
+public struct MalformedURLError: Error {
```
cobbal, line 12: Pretty sure I saw this earlier in some other file

Tests/FishyJoesExecuteTests/SwiftPackageTests.swift
```diff
@@ -29,3 +29,58 @@ class SwiftPackageTests: XCTestCase {
         XCTAssertEqual(parsed.dependencyMap["crigeo"], URL(string: "https://github.com/cricut/CriGeo"))
     }
 }
+/*
```
cobbal: Is this comment valuable?

node-runtime/fishyjoes-runtime-wasm/Runtime.js
cobbal, line 1: These files all look the same. Should we just commit one copy like we do for the .d.ts?

cobbal (changes requested):

.github/workflows/integration-tests.yaml
```diff
@@ -77,47 +77,49 @@ jobs:
...
+      - name: Test kotlin from gradle via fishy-joes
         shell: zsh {0}
         run: |
           set -euxo pipefail
 
           cd integration-tests/TestAPI-bindings
-          swift run -- fishy-joes --dart build test
+          swift run -- fishy-joes --kotlin-fast build test
+
+  #TODO: integration-tests-windows, but right now, windows is too flaky about not being able to call swift-build.exe
```
cobbal: Integration aren't trying to test every platform, just the platforms we need to cover the different cases. None of our processes run only on windows, so windows doesn't belong here

.github/workflows/integration-tests.yaml
```diff
@@ -224,6 +226,8 @@ jobs:
           mkdir -p $libdir
           ./scripts/compile-kotlin-native-runtime.sh
 
+      # Running this step messes with the environment variables significantly,
```
cobbal: It's not environment variables it's permissions.

.github/workflows/node-runtime.yaml
```diff
@@ -36,11 +38,49 @@ jobs:
...
+          node-version: '16.x'
+          registry-url: 'https://npm.pkg.github.com/'
+          scope: '@cricut'
+
+      - name: NPM Pack
+        shell: zsh {0}
+        run: |
+          set -euxo pipefail
+
+          cat ${{ env.LIB_DIR }}/package.json | jq '.version = "${{ env.VERSION }}"' > ${{ env.LIB_DIR }}/package.json.upd
```
cobbal: When in the shell, prefer using the shell to expand environment variables (e.g. `$LIB_DIR`) for fewer escaping errors

cobbal: For jq, use it's env variable accessor `.version = env.VERSION` for similar reasons. https://jqlang.github.io/jq/manual/#$env-env

Sources/FishyJoesCore/Unparse/TypeScriptAnnotations.swift
```diff
@@ -260,7 +260,7 @@ extension TypeScriptAnnotations {
         let fragment = SourceFragment(sourceryDestination: "file:NodeInterface/\(defaultNamespace).d.ts")
 
         fragment.output("import { Runtime } from '@cricut/fishyjoes-runtime';")
-        fragment.output("export * from '@cricut/fishyjoes-runtime';")
+        fragment.output("export { Runtime } from '@cricut/fishyjoes-runtime';")
```
cobbal, line 263: I think this makes the previous line redundant? Might need to as Kelly

Sources/FishyJoesExecute/CodeGen.swift
```diff
@@ -645,6 +633,30 @@ extension CodeGen {
             for platform in platforms {
                 switch platform {
                 case .wasm, .node:
+                    // Update version number in package, if provided
+                    if let version = version {
+                        // Read package.json and update the version
+                        let packageJsonPath = "\(platform.outputDir(config))/package.json"
```
cobbal, line 639: We should not be reading output, only writing. I'm not sure what this whole section is trying to accomplish

Sources/FishyJoesExecute/Resources/bindings-template/.github/workflows/c-sharp.yaml
```diff
@@ -16,9 +15,13 @@ concurrency:
...
+  GITHUB_TOKEN: ${{ secrets.CRICUT_GPR_TOKEN }}
+  GITHUB_USER: cricut-builduser
+
 jobs:
   build-test-c-sharp-mac:
-    name: "mac: Build and test C# bindings"
+    name: "macOS: Build and test C# bindings"
     runs-on: macOS-12
     env:
       GITHUB_TOKEN: ${{ secrets.CRICUT_GPR_TOKEN }}
```
cobbal: If you set this for the entire workflow, delete the redundant per-job ones

Sources/FishyJoesExecute/Resources/bindings-template/c-sharp/Cricut.__MODULE_NAME__.Tests/ExampleTests.cs
```diff
@@ -2,6 +2,6 @@
 using Xunit;
 
 namespace Cricut.__MODULE_NAME__.Tests {
-    public class UnitTest1 {
```
cobbal, line 5: UnitTest1 is the name the official microsoft tools created. Let them live with their poor choices. (kidding... mostly...)

Sources/FishyJoesExecute/Resources/bindings-template/dart/.gitignore
```diff
@@ -10,5 +10,5 @@ build/
 pubspec.lock
 
 # generated code/libs
-*.freezed.dart
```
cobbal, line 13: I \*really\* don't like this as a long term solution

dart-runtime/bin/setup.dart
```diff
@@ -60,7 +60,7 @@ void main() async {
           io.exit(1);
         }
         final repoName = pathMatch[1]!;
-        downloads.add(Download(repoName, ref, "dart-binaries.tgz"));
```
cobbal, line 63: I could have sworn I've committed this change before...

cobbal (changes requested):
Looking pretty good. A few minor shell-related issues

PR #148, opened 2023-12-08
Node.js Windows Support

mstokercricut:
* Adds Windows 10 as a supported target platform via Node.js and TypeScript
* Changes how Wasm module registration is done to fit better with how it must be done for Node.js
* Provides several improvements to the Platform type for use in code generation


Sources/FishyJoesExecute/Resources/bindings-template/Package.swift
```diff
@@ -45,8 +45,8 @@ let package = Package(
         ),
         .package(
             // NOTE: Must reference releases using "branch" instead of "exact" because of required usage of "unsafeFlags" in FishyJoes
-            // url: "https://github.com/cricut/FishyJoes", .branch(fishyJoesVersion)
-            path: "../FishyJoes"
+            url: "https://github.com/cricut/FishyJoes", .branch(fishyJoesVersion)
```
cobbal, line 48: This is a very annoying change, as any time the template is run "swift run fishy-joes" will never work. Before it would sometimes work (especially in development)

.github/workflows/integration-tests.yaml
```diff
@@ -263,6 +257,104 @@ jobs:
...
+
+          ./scripts/compile-node-runtime.sh
+
+      - name: Test TypeScript from npm via fishy-joes
+        shell: bash
+        run: |
+          set -euxo pipefail
+
+          cd integration-tests/TestAPI-bindings
+          # TODO: Wasm build support on Windows (needs SwiftWasm compiler)
```
cobbal: Not really a TODO. We don't test wasm on mac either. We don't need to test it on every platform, as that's testing that's a part of swift-wasm, not part of this repo.

.github/workflows/integration-tests.yaml
```diff
@@ -263,6 +257,104 @@ jobs:
...
+          #$SWIFT run -- fishy-joes --nodejs --wasm build test
+          $SWIFT run -- fishy-joes --nodejs build test
```
cobbal: might be best practice to quote `"$SWIFT"`. Windows is known for putting spaces in program paths, even if the swift package doesn't

.github/workflows/integration-tests.yaml
```diff
@@ -263,6 +257,104 @@ jobs:
...
+
+      - name: Test Dart from dart via fishy-joes
+        shell: bash
+        run: |
+          set -euxo pipefail
+
+          cd integration-tests/TestAPI-bindings
+          # TODO: TestAPI.dll is referenced from TestAPI-iota.dll, but for some reason
+          #       it cannot be found in windows/native and must be in the working directory
+          #$SWIFT run -- fishy-joes --dart build test
```
cobbal: https://github.com/cricut/CriGeo-bindings/blob/main/.github/workflows/dart.yaml#L161-L162 (powershell)

No need to put it in the current directory, just make sure the Path includes it

.github/workflows/iota-runtime.yaml
```diff
@@ -137,26 +137,23 @@ jobs:
...
+          ./scripts/compile-iota-runtime.sh
 
       - name: Copy swift runtime DLLs
         shell: bash
         run: |
           set -euxo pipefail
 
-          cp -R /c/Library/icu-*/usr/bin/*.dll $LIB_DIR
-          cp -R /c/Library/Swift-development/bin/*.dll $LIB_DIR
+          mkdir -p $LIB_DIR
```
cobbal: quote LIB_DIR here if it's quoted in the next line for consistency.

Package.swift
```diff
@@ -56,14 +65,14 @@ let package = Package(
...
             D.package(url: "https://github.com/apple/swift-argument-parser", from: "1.2.2"),
         ]
     ) + (androidCompatibleOnly || wasmCompatibleOnly ? [] : [
         D.package(url: "https://github.com/jpsim/Yams", .upToNextMinor(from: "5.0.3")),
     ]),
     targets: [
-        T.systemLibrary(name: "NodeAPI"),
+        // Kotlin / Java
         T.systemLibrary(name: "JNI"),
         T.target(name: "FishyJoesCommonRuntime"),
```
cobbal: This one is not java-specific

Package.swift
```diff
@@ -249,13 +264,31 @@ let package = Package(
...
+                    .linkedLibrary("node.lib"),
+                    .linkedLibrary("delayimp.lib", .when(platforms: [.windows])),
+                    // Relative reference when building from FishyJoes directory
+                    .unsafeFlags(["-Xlinker", "/LIBPATH:.\\Sources\\NodeAPIResolve\\lib\\windows.x86_64"], .when(platforms: [.windows])),
+                    // Relative reference when building from FishyJoes/integration-tests/TestAPI-bindings directory
+                    .unsafeFlags(["-Xlinker", "/LIBPATH:..\\..\\Sources\\NodeAPIResolve\\lib\\windows.x86_64"], .when(platforms: [.windows])),
+                    // Relative reference when building from FishyJoes/integration-tests/TestAPI-bindings/Sources/Generated/NodeNativeShim directory
+                    .unsafeFlags(["-Xlinker", "/LIBPATH:..\\..\\..\\..\\..\\Sources\\NodeAPIResolve\\lib\\windows.x86_64"], .when(platforms: [.windows])),
+                    // Relative reference when building from a bindings repo directory
+                    .unsafeFlags(["-Xlinker", "/LIBPATH:.\\.build\\checkouts\\FishyJoes\\Sources\\NodeAPIResolve\\lib\\windows.x86_64"], .when(platforms: [.windows])),
```
cobbal: This feels incredibly fraglie

Sources/FishyJoesCore/Translators/CSharpTranslator.swift
```diff
@@ -225,20 +225,6 @@ final class CSharpTranslator: Translator {
...
-        }
-        switch value {
-        case "nil":
-            return "null"
-        case "true", "false":
-            return value
-        default:
-            return nil
-        }
+        return value == "nil" ? "null" : literal(value: value)
```
cobbal: I really don't like this being centralized. Essentially it's asserting that any valid swift code is valid C# and JS and kotlin.

Sources/FishyJoesCore/Translators/NodeTranslator.swift
```diff
@@ -321,14 +303,54 @@ struct NodeTranslator: Translator {
             }
             nodeTypeListFragment.output("return exports")
         }
+        nodeTypeListFragment.blankLine()
+        nodeTypeListFragment.output("@available(*, deprecated, message: \"Not actually deprecated, but this silences warnings because it may refer to deprecated methods\")")
+        nodeTypeListFragment.output("@_cdecl(\"registerModule\(context.module.name)\")")
+        nodeTypeListFragment.outputBlock("public func registerModule\(context.module.name)C(env: napi_env, exports: napi_value) -> napi_value? {") {
```
cobbal: nit: It's really hard to see the name difference. Putting it at the front would be more obvious: "cRegisterModuleCriGeo"

cobbal: Or should this one be "registerModuleCriGeo" and the other one just "CriGeo_NodeInterface.registerModule" since it doesn't need a cdecl symbol?

None of this really matters much I suppose

Sources/FishyJoesCore/Translators/NodeTranslator.swift
```diff
@@ -321,14 +303,54 @@ struct NodeTranslator: Translator {
...
 
         let exportFragment = SourceFragment(sourceryDestination: "file:NodeInterface/@_exported.swift")
         exportFragment.output("@_exported import \(context.module.name)")
         for dependency in context.module.dependencies {
             exportFragment.output("@_exported import \(dependency)_NodeInterface")
         }
 
-        return [nodeTypeListFragment, exportFragment]
+        let nodeNativeShimFragment = SourceFragment(sourceryDestination: "file:NodeNativeShim/NAPIRegisterModule.c")
+        nodeNativeShimFragment.output("#include <node_api.h>")
```
cobbal, line 324: Could benefit from outputting some blank lines in here. The generated file looks dense and hard to read

Sources/NodeAPIResolve/win_delay_load_hook.cc
```diff
@@ -0,0 +1,41 @@
...
+#ifndef WIN32_LEAN_AND_MEAN
+#define WIN32_LEAN_AND_MEAN
+#endif
+
+#include <windows.h>
+
+#include <delayimp.h>
+#include <string.h>
+
+#define HOST_BINARY "node.exe"
```
cobbal: Is this the name of the binary in design space? Could be an issue.

documentation/FishyJoes Hello World.txt
cobbal, line 1: This is very rough documentation. A couple ideas for improvement:

 - use markdown
 - put some context in, instead of a raw terminal dump
 - please no spaces in file names
 - even I have no idea what you're referring to under "Temporary Fixes"

integration-tests/TestAPI-bindings/node-test/AttributedStrings.test.ts
```diff
@@ -4,6 +4,17 @@ const AttributedString = Runtime.AttributedString
...
+test('Crashy', () => {
+    const attributedString = AttributedStrings.polyglot
+    // const attributedString = AttributedString.createJoining([
+    //     AttributedStrings.polyglot,
+    //     AttributedString.create("")
+    // ])
+    // expect(attributedString.string).toEqual("Hello Olá こんにちは")
+
+    const index =  attributedString.characters.indexAfter(attributedString.startIndex)
+})
```
cobbal: Is this supposed to be committed?

integration-tests/TestAPI-bindings/node-test/package.json
```diff
@@ -7,8 +7,11 @@
...
     "test-native-ubuntu": "node --force-node-api-uncaught-exceptions-policy --experimental-vm-modules node_modules/jest/bin/jest.js --run-in-band --detect-open-handles --force-exit --debug --no-watchman -c jest.config.native-ubuntu.cjs",
+    "test-native-ubuntu-lldb": "lldb -- node --force-node-api-uncaught-exceptions-policy --experimental-vm-modules node_modules/jest/bin/jest.js --run-in-band --detect-open-handles --force-exit --debug --no-watchman -c jest.config.native-ubuntu.cjs",
     "test-native-windows": "node --force-node-api-uncaught-exceptions-policy --experimental-vm-modules node_modules/jest/bin/jest.js --run-in-band --detect-open-handles --force-exit --debug --no-watchman -c jest.config.native-windows.cjs",
+    "test-native-windows-lldb": "lldb -- node --force-node-api-uncaught-exceptions-policy --experimental-vm-modules node_modules/jest/bin/jest.js --run-in-band --detect-open-handles --force-exit --debug --no-watchman -c jest.config.native-windows.cjs",
```
cobbal, line 14: Good additions, but this file is becoming very terrible. I wonder if there's a way to abstract the common options.

integration-tests/TestAPI-bindings/node-test/package.json
```diff
@@ -22,12 +25,12 @@
     "@cricut/fishyjoes-runtime-native-windows": "file:../../../node-runtime/fishyjoes-runtime-native-windows",
     "@cricut/fishyjoes-runtime-native-macos": "file:../../../node-runtime/fishyjoes-runtime-native-macos",
     "@cricut/fishyjoes-runtime-native-ubuntu": "file:../../../node-runtime/fishyjoes-runtime-native-ubuntu",
-    "@cricut/testapi-native-windows": "file:../output/node-native-windows",
     "@cricut/testapi-native-macos": "file:../output/node-native-macos",
     "@cricut/testapi-native-ubuntu": "file:../output/node-native-ubuntu",
     "@cricut/testapi-wasm": "file:../output/wasm"
   },
   "dependencies": {
+    "@cricut/testapi-native-windows": "file:../output/node-native-windows",
```
cobbal, line 33: Why is windows special? Or was this a stray "npm install" that caused this?

node-runtime/fishyjoes-runtime-native-macos/Runtime.js
```diff
@@ -1,10 +1,10 @@
 import { createRequire } from 'module';
 const require = createRequire(import.meta.url);
-const { Runtime } = require('./Runtime.cjs');
+const { Optional, Runtime } = require('./Runtime.cjs');
```
cobbal: Does optional exist in javascript? I thought it was just a type.

node-runtime/fishyjoes-runtime-native-ubuntu/package.json
```diff
@@ -7,8 +7,7 @@
   "description": "FishyJoes Runtime for Node.js on Ubuntu",
   "main": "Runtime.js",
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1",
-    "postinstall": "./postinstall.sh"
```
cobbal, line 11: yay!

scripts/compile-node-runtime.sh
```diff
@@ -10,17 +10,47 @@ fi
 CONFIGURATION="${CONFIGURATION:-release}"
 SKIP_LIPO="${SKIP_LIPO:-0}"
 
-if [[ $(uname -s) == "Darwin" && $SKIP_LIPO == "0" ]]; then
+SHIM_DIR="Sources/FishyJoesNodeRuntime_NodeNativeShim"
```
cobbal: This is not a source of FishyJoes/package.swift, it probably doesn't belong in Sources. Probably should go in node-runtime instead?

cobbal (changes requested):

Sources/FishyJoesCore/Translators/CSharpTranslator.swift
mstokercricut: It does? It looks like it translates "nil" to "null", then tries to translate the value as a literal, which all our languages share a common definition for. If that fails, it reports that the value could not be translated.

Am I missing something? Potentially Swift's string conversion of a double value may not work in C# as a literal in its compiler, but I think that they all do...

Sources/NodeAPIResolve/win_delay_load_hook.cc
mstokercricut: Ya, it could be. If it isn't "node.exe" somewhere in the path, though, how would we ever be able to find it reliably?

integration-tests/TestAPI-bindings/node-test/AttributedStrings.test.ts
mstokercricut: Nope.

integration-tests/TestAPI-bindings/node-test/package.json
mstokercricut: If npm can call back into npm, we could do that. E.g.

"test-native-windows-lldb": "lldb -- npm run test-native-windows"

I figured that npm might actually then call out to node, so lldb would then miss the part we actually want debugged. In fact, on Windows, it almost certainly does this. Without generating package.json, adding this without duplication seems impossible to me. Maybe adding some environment variables or something, but that seems worse...

node-runtime/fishyjoes-runtime-native-macos/Runtime.js
mstokercricut: Not that I can find. You actually added the original definition for `Optional`.

node-runtime/fishyjoes-runtime-native-ubuntu/package.json
mstokercricut: I wish this could actually go away. The definition just moved to CodeGen.swift line 600. For safety, I had macOS do the symlink too.

documentation/FishyJoes Hello World.txt
mstokercricut: I'll toy with it a bit more before including it

Package.swift
mstokercricut: I agree. There must be a better way. Require that on Windows a dev adds Node to the system lib path?

By moving to using clang to compile the shim, 1 of these went away at least.

node-runtime/fishyjoes-runtime-native-macos/Runtime.js
mstokercricut: Removed! Annoying to track the js and ts exports separately, but done.

Package.swift
```diff
@@ -8,6 +8,7 @@ let disableGeneration = env["DISABLE_GENERATION"] == "1"
 let wasmCompatibleOnly = env["WASM_ONLY"] == "1"
 let androidCompatibleOnly = env["ANDROID_COMPATIBLE_ONLY"] == "1"
 let javaHome = env["JAVA_HOME_11_X64"] ?? env["JAVA_HOME"]
+let extraLibPath = env["EXTRA_LIBPATH"]?.split(separator:";") ?? []
```
cobbal: lint: space after ":"

cobbal: Would you mind also adding Package.swift to the "included" section of swiftlint.yaml?

Package.swift
```diff
@@ -96,16 +89,26 @@ let package = Package(
...
+        T.target(
+            name: "NodeAPI",
+            linkerSettings: [
+                // On Windows, node.lib must be in the LIB environment variable when building FishyJoes, the runtime, or bindings libraries, via "swift build"
+                // Use the EXTRA_LIBPATH environment variable to cause "swift build" to add paths to LIB when envoked
+                // When invoked via "fishy-joes build", node.lib will be downloaded and put in a directory accessible by the LIB environment variable as part of the build process
+                .linkedLibrary("node.lib", .when(platforms: [.windows])),
+                .linkedLibrary("delayimp.lib", .when(platforms: [.windows])),
+                .unsafeFlags(
+                    Array(extraLibPath.map { ["-Xlinker", "/LIBPATH:\($0)"] }.joined()),
```
cobbal: Just use flatMap here

Package.swift
```diff
@@ -96,16 +89,26 @@ let package = Package(
             ]
         ),
         // TypeScript Node.js / Wasm
-        T.systemLibrary(name: "NodeAPI"),
+        T.target(
+            name: "NodeAPI",
+            linkerSettings: [
+                // On Windows, node.lib must be in the LIB environment variable when building FishyJoes, the runtime, or bindings libraries, via "swift build"
```
cobbal: We're not actually using LIB env var, are we? This is all being passed by flags and differently named vars

Sources/FishyJoesExecute/DownloadNodeLib.swift
```diff
@@ -0,0 +1,37 @@
...
+
+    public mutating func run() throws {
+        ExternalCommand.verbose = true
+        try Self.download(url: url, destination: destination, force: force)
+    }
+
+    static public func download(
+        url: String? = nil,
+        destination: String? = nil,
+        force: Bool? = nil
```
cobbal: nit: force doesn't need to be optional

cobbal (approved):
This is looking very solid.

Package.swift
mstokercricut: 👍🏿

PR #150, opened 2023-12-14
Update README.md

mstokercricut:


README.md
```diff
@@ -24,6 +24,41 @@ The recommended way to ensure you have everything you will need for any FishyJoe
 1. [What FishyJoes Can and Cannot Do](documentation/cans-and-cannots.md)
 2. Was that a question?
    **No, but it was frequently asked**
+3. What annotations can be made to Swift code to mark it for FishyJoes to process?
+```
+// Export a type member-wise, including all its properties (used for simple data objects, data lives on foreign side)
```
cobbal: Export is more than a struct option, and structs are the only case where "member-wise" really describes what's going on. Export means "export normally and transparently". For structs that's memberwise, for enums it's case-wise, for classes it's as a reference. exportReference means "force export as a reference" and is just as valid on enums as it is on structs

README.md
```diff
@@ -24,6 +24,41 @@ The recommended way to ensure you have everything you will need for any FishyJoe
...
+    /// <!-- FishyJoes.export(someGenericMethod, generic: [G: [SomeConcreteSwiftTypeSatisfyingG]) -->
+    public func someGenericMethod<G: Generic>(parameter: G) { \* ... *\ }
+
+    // Export a method omitting one or more parameters that have default values from the foreign interface
+    /// <!-- FishyJoes.export(someMethodWithOptionals, omitParameters: [someDefaultedParameter]) -->
+    public func someMethodWithOptionals(parameter: SomeOtherSwiftType someDefaultedParameter: Int? = nil) { \* ... *\ }
+
+    // Export a method, then add parameters with defaults later, without breaking interface
+    /// <!-- FishyJoes.export(someMethodThatChanges, compatibilityOrder: [newParameterTwo]) -->
+    public func someMethodThatChanges(parameterOne: Int, newParameterTwo: SomeOtherSwiftType? = nil) { \* ... *\ }
```
cobbal: Missing a few annotations like "isOverride", "cSharp", and "noReturn". (Didn't even remember that last one was in there). The way this FAQ is stated sounds like it's a complete list. Even if we don't go into detail on them, we should probably mention they exist.

cobbal (changes requested):

cobbal (approved):

PR #161, opened 2024-01-12
Bscothern/async -> main

cobbal:


Sources/FishyJoesCore/TranslatedTypes/TranslatedType.swift
```diff
@@ -52,6 +52,8 @@ extension TranslatedType {
             return .void
         } else if let unsignedPrimitiveType = self as? TranslatedUnsignedPrimitive {
             return .unsigned(jniType: unsignedPrimitiveType.jniType)
+        } else if let functionType = self as? TranslatedFunction {
+            return .function(signature: functionType.kotlinName, parmeters: functionType.parameters.map(\.kotlinType))
```
cobbal: What happens to the async bool here?

Sources/FishyJoesCore/Translators/KotlinTranslator.swift
```diff
@@ -36,7 +38,16 @@ final class KotlinTranslator: Translator {
...
             let resolved = context.resolve(type: parameter.type, generics: exportAnnotation.genericOverrides)
             return (name: parameter.name, type: resolved.converterType.name + ".CType")
-        }
+        } + {
+            guard method.isAsync else {
+                return []
+            }
+            let returnType = context.resolve(type: method.returnType, generics: exportAnnotation.genericOverrides).converterType.name
+            return [
+                (name: "_successContinuation", type: "Function1Converter<\(returnType)\(returnType.hasSuffix("Converter") ? "" : ".CType"), VoidConverter>.CType"),
```
cobbal: This is always the converter type, returnType should probably be returnConverterType, and should always have .CType (as that's part of the protocol)

Sources/FishyJoesCore/Translators/KotlinTranslator.swift
```diff
@@ -62,27 +76,98 @@ final class KotlinTranslator: Translator {
...
+                            }
+                            fragment.output("try? _successContinuationRef.destroy()")
+                            fragment.output("try? _failureContinuationRef.destroy()")
+                        }
+                        fragment.outputBlock("do {", closeWith: "}", newLineTerminated: false) {
+                            fragment.outputMap(method.parameters, separator: "\n") { parameter in
+                                let resolved = context.resolve(type: parameter.type, generics: exportAnnotation.genericOverrides)
+                                return "let \(parameter.name) = try \(resolved.converterType.name).fromJava(object: \(parameter.name)Ref.createLocalRef(env: _javaEnv), env: _javaEnv)"
+                            }
+                            fragment.output("\(method.isMutating ? "var" : "let") value: Result<\(method.returnType.name), any Error>\(method.isMutating ? "!" : "")")
```
cobbal: I think this can just be let?

Sources/FishyJoesJavaRuntime/Env.swift
```diff
@@ -51,6 +54,69 @@ public struct Env {
...
+            Thread.current.threadDictionary.setValue(jvmAttachCount + 1, forKey: Env.jvmAttachCountKey)
+            return rawJavaEnvPointer
+                .map(OpaquePointer.init)
+                .map(UnsafeMutablePointer.init)
+                .map(Env.init(env:))!
+        case JNI_EDETACHED:
+            Thread.current.threadDictionary.setValue(1, forKey: Env.jvmAttachCountKey)
+            var javaEnv: UnsafeMutablePointer<JNIEnv?>?
+            let attachStatus = jvmInvoke.AttachCurrentThread(jvm, &javaEnv, nil)
+            guard attachStatus == JNI_OK else {
```
cobbal, line 81: Should the thread-local count be 0 if an error happens here?

Sources/FishyJoesJavaRuntime/Env.swift
```diff
@@ -51,6 +54,69 @@ public struct Env {
...
+    public static func aquireJVMThread(on jvm: UnsafeMutablePointer<JavaVM?>?) throws -> Env {
+        var rawJavaEnvPointer: UnsafeMutableRawPointer?
+        guard let jvmInvoke = jvm?.pointee?.pointee else {
+            throw JNIError(message: "Invalid JVM pointer")
+        }
+        // This is the only way to determine if the current thread is already attached.
+        let getEnvStatus = jvmInvoke.GetEnv(jvm, &rawJavaEnvPointer, JNI_VERSION_1_4)
+        switch getEnvStatus {
+        case JNI_OK:
+            let jvmAttachCount = Thread.current.threadDictionary.value(forKey: Env.jvmAttachCountKey) as? Int ?? 1
```
cobbal, line 72: Could use a comment like "The JVM is attached to this thread by something other than FishyJoes, start the ref count at 1 so that we don't detach a thread we don't own"

Sources/FishyJoesJavaRuntime/Env.swift
```diff
@@ -51,6 +54,69 @@ public struct Env {
...
+              let jvmAttachCount = Thread.current.threadDictionary.value(forKey: Env.jvmAttachCountKey) as? Int else {
+            return
+        }
+        if jvmAttachCount == 1 {
+            let detachStatus = jvmInvoke.DetachCurrentThread(jvm)
+            guard detachStatus == JNI_OK else {
+                throw JNIError(message: "Failed to detach current thread from JVM \(detachStatus)")
+            }
+        }
+        Thread.current.threadDictionary.setValue(jvmAttachCount - 1, forKey: Env.jvmAttachCountKey)
```
cobbal: This should probably be deferred

Sources/FishyJoesCore/Unparse/KotlinClass.swift
```diff
@@ -126,12 +132,13 @@ class KotlinClass: NestedClass {
...
                     fragment.output("override ", newLineTerminated: false)
                 }
-                fragment.outputBlock("fun \(method.name)(", newLineTerminated: false) {
+                fragment.outputBlock("\(method.isSuspend ? "suspend " : "")fun \(method.name)(", newLineTerminated: false) {
                     let filteredParameters = method.parameters.filter { !excludedCompatibilityParameters.contains($0.name) }
                     fragment.outputMap(filteredParameters, separator: ",") { parameter in
                         let labelComment = parameter.labelComment.map { "/* \($0) */ " } ?? ""
                         let defaultValue = parameter.defaultValue.map { " = \($0)" } ?? ""
-                        return "\(labelComment)\(parameter.name): \(parameter.type.kotlinType)\(defaultValue)"
+                        let kotlinType: String = (method.isSuspend && parameter.type.isFunction) ? "(suspend \(parameter.type.kotlinType.dropFirst())" : parameter.type.kotlinType
```
mstokercricut: A comment like "Rather than allowing a suspend function to take in another non-suspend function, we have decided to explicitly disallow that configuration so this does not need to be recursive"

Sources/FishyJoesCore/Unparse/KotlinClass.swift
```diff
@@ -140,6 +147,58 @@ class KotlinClass: NestedClass {
...
+                                                        parametersToForward.append("p\(parameterIndex)")
+                                                    }
+                                                default:
+                                                    break
+                                                }
+
+                                                fragment.outputBlock("{\(parametersToForward)\(parametersToForward.isEmpty ? "" : " ->")", closeWith: "}", newLineTerminated: false) {
+                                                    fragment.outputBlock("try {", newLineTerminated: false) {
+                                                        fragment.outputBlock("CoroutineScope(Dispatchers.Default).future {", newLineTerminated: false) {
+                                                            // TODO: Forward parameters
```
mstokercricut: Nuke this?

Sources/FishyJoesCore/Unparse/KotlinClass.swift
```diff
@@ -140,6 +147,58 @@ class KotlinClass: NestedClass {
...
                     precondition(compatibilityParameters.isEmpty, "internal error: compatibilityOrder can't be used with a non-native body")
                     fragment.output(" = \(body)\(method.returnType.toKotlinType)")
+                } else if method.isSuspend {
+                    fragment.outputBlock(" {") {
+                        fragment.outputBlock("return coroutineScope {") {
+                            fragment.outputBlock("async {", newLineTerminated: false) {
+                                fragment.outputBlock("suspendCancellableCoroutine { continuation: CancellableContinuation<\(method.returnType.kotlinType)\(method.returnType.kotlinType == "kotlin.Unit" ? "?" : "")> ->", closeWith: "}") {
+                                    fragment.outputBlock("__jni_\(method.name)(", closeWith: ")", newLineTerminated: false) {
+                                        for parameter in method.parameters {
+                                            if parameter.type.isFunction {
```
mstokercricut: Why are we looking at function parameters directly? This should compose nicely if dispatched to the type (AsyncFunctionConverter) to handle...

Sources/FishyJoesCore/Unparse/KotlinClass.swift
```diff
@@ -159,11 +218,22 @@ class KotlinClass: NestedClass {
...
             fragment.output("@JvmName(\"__jni_\(method.name)\")")
             fragment.outputBlock("private external fun __jni_\(method.name)(", newLineTerminated: false) {
-                fragment.outputMap(method.parameters, separator: ",") { parameter in
+                fragment.outputMap(
+                    method.parameters + {
+                        guard method.isSuspend else {
+                            return []
+                        }
+                        return [
+                            (labelComment: nil, name: "successContinuation", type: .named(package: nil, name: "(\(method.returnType.kotlinType)\(method.returnType.kotlinType == "kotlin.Unit" ? "?" : "")) -> Unit"), defaultValue: nil),
```
mstokercricut: Put tuple arguments on their own lines?

Sources/FishyJoesCore/Unparse/KotlinClass.swift
```diff
@@ -177,6 +247,10 @@ extension KotlinClass.KType: CustomStringConvertible {
         "FIXME: You should not use this, you should use one of the representations below. \(jvmType)"
     }
 
+    var isFunction: Bool {
+        kotlinType.contains("->")
```
mstokercricut: Change to switch on self, and return true in `function` case.

Sources/FishyJoesCore/Unparse/KotlinClass.swift
```diff
@@ -310,6 +385,7 @@ class KotlinEnumClass: KotlinClass {
 
     override func output(to fragment: SourceFragment) {
         document(documentation, fragment: fragment)
+        fragment.output("@OptIn(ExperimentalCoroutinesApi::class)")
```
mstokercricut, line 332: Should only output this for async? Or just on everything?

Sources/FishyJoesJavaRuntime/JavaFunction.swift
```diff
@@ -188,18 +337,69 @@ extension Function1Converter: JavaConverter where R: JavaConverter, P0: JavaConv
...
+        }
+    }
+
+    public static func toJava(_ value: @escaping SwiftType, env: Env) throws -> jobject? {
+        let erased = AnyAsyncFunction1 { env, p0 in
+            let v0 = try P0.fromJava(object: p0, env: env)
+            return try await R.toJavaObject(value(v0), env: env)
+        }
+        let ptr = jvalue(pointer: Box(erased).retainedOpaque())
+        return try env.NewObject(SwiftAsyncFunctionImpl.implClass, SwiftAsyncFunctionImpl.constructor, jvalue(i: 0), ptr)
```
mstokercricut: Arity of 1?

Sources/FishyJoesJavaRuntime/JavaFunction.swift
```diff
@@ -234,18 +434,72 @@ extension Function2Converter: JavaConverter where R: JavaConverter, P0: JavaConv
...
+    }
+
+    public static func toJava(_ value: @escaping SwiftType, env: Env) throws -> jobject? {
+        let erased = AnyAsyncFunction2 { env, p0, p1 in
+            let v0 = try P0.fromJava(object: p0, env: env)
+            let v1 = try P1.fromJava(object: p1, env: env)
+            return try await R.toJavaObject(value(v0, v1), env: env)
+        }
+        let ptr = jvalue(pointer: Box(erased).retainedOpaque())
+        return try env.NewObject(SwiftAsyncFunctionImpl.implClass, SwiftAsyncFunctionImpl.constructor, jvalue(i: 0), ptr)
```
mstokercricut: Arity of 2?

Sources/FishyJoesJavaRuntime/JavaFunction.swift
```diff
@@ -283,18 +537,75 @@ extension Function3Converter: JavaConverter where R: JavaConverter, P0: JavaConv
...
+
+    public static func toJava(_ value: @escaping SwiftType, env: Env) throws -> jobject? {
+        let erased = AnyAsyncFunction3 { env, p0, p1, p2 in
+            let v0 = try P0.fromJava(object: p0, env: env)
+            let v1 = try P1.fromJava(object: p1, env: env)
+            let v2 = try P2.fromJava(object: p2, env: env)
+            return try await R.toJavaObject(value(v0, v1, v2), env: env)
+        }
+        let ptr = jvalue(pointer: Box(erased).retainedOpaque())
+        return try env.NewObject(SwiftAsyncFunctionImpl.implClass, SwiftAsyncFunctionImpl.constructor, jvalue(i: 0), ptr)
```
mstokercricut: Arity of 3?

Sources/FishyJoesJavaRuntime/JavaFunction.swift
```diff
@@ -335,18 +646,78 @@ extension Function4Converter: JavaConverter where R: JavaConverter, P0: JavaConv
...
+    public static func toJava(_ value: @escaping SwiftType, env: Env) throws -> jobject? {
+        let erased = AnyAsyncFunction4 { env, p0, p1, p2, p3 in
+            let v0 = try P0.fromJava(object: p0, env: env)
+            let v1 = try P1.fromJava(object: p1, env: env)
+            let v2 = try P2.fromJava(object: p2, env: env)
+            let v3 = try P3.fromJava(object: p3, env: env)
+            return try await R.toJavaObject(value(v0, v1, v2, v3), env: env)
+        }
+        let ptr = jvalue(pointer: Box(erased).retainedOpaque())
+        return try env.NewObject(SwiftAsyncFunctionImpl.implClass, SwiftAsyncFunctionImpl.constructor, jvalue(i: 0), ptr)
```
mstokercricut: Arity of 4?

Sources/FishyJoesJavaRuntime/JavaFunction.swift
```diff
@@ -390,18 +761,81 @@ extension Function5Converter: JavaConverter where R: JavaConverter, P0: JavaConv
...
+        let erased = AnyAsyncFunction5 { env, p0, p1, p2, p3, p4 in
+            let v0 = try P0.fromJava(object: p0, env: env)
+            let v1 = try P1.fromJava(object: p1, env: env)
+            let v2 = try P2.fromJava(object: p2, env: env)
+            let v3 = try P3.fromJava(object: p3, env: env)
+            let v4 = try P4.fromJava(object: p4, env: env)
+            return try await R.toJavaObject(value(v0, v1, v2, v3, v4), env: env)
+        }
+        let ptr = jvalue(pointer: Box(erased).retainedOpaque())
+        return try env.NewObject(SwiftAsyncFunctionImpl.implClass, SwiftAsyncFunctionImpl.constructor, jvalue(i: 0), ptr)
```
mstokercricut: Arity of 5?

Sources/FishyJoesJavaRuntime/JavaFunction.swift
```diff
@@ -447,3 +881,70 @@ extension Function6Converter: JavaConverter where R: JavaConverter, P0: JavaConv
...
+            let v0 = try P0.fromJava(object: p0, env: env)
+            let v1 = try P1.fromJava(object: p1, env: env)
+            let v2 = try P2.fromJava(object: p2, env: env)
+            let v3 = try P3.fromJava(object: p3, env: env)
+            let v4 = try P4.fromJava(object: p4, env: env)
+            let v5 = try P5.fromJava(object: p5, env: env)
+            return try await R.toJavaObject(value(v0, v1, v2, v3, v4, v5), env: env)
+        }
+        let ptr = jvalue(pointer: Box(erased).retainedOpaque())
+        return try env.NewObject(SwiftAsyncFunctionImpl.implClass, SwiftAsyncFunctionImpl.constructor, jvalue(i: 0), ptr)
```
mstokercricut: Arity of 6?

Sources/FishyJoesJavaRuntime/JavaReference.swift
```diff
@@ -2,23 +2,41 @@ import FishyJoesCommonRuntime
...
+    public private(set) var object: jobject?
+    private let vm: UnsafeMutablePointer<JavaVM?>?
 
-    init(local: jobject?, env: Env) throws {
+    public init(local: jobject?, env: Env) throws {
         self.object = try env.globalRef(local)
         vm = try env.GetJavaVM()
     }
 
+    public convenience init<T: JavaConverter>(local: T, env: Env) throws where T.SwiftType == T, T.CType == T {
```
mstokercricut: Is this something we need?

Sources/FishyJoesJavaRuntime/JavaReference.swift
```diff
@@ -2,23 +2,41 @@ import FishyJoesCommonRuntime
...
+    }
+
+    public func destroy() throws {
         if let object = object {
-            var envRaw: UnsafeMutableRawPointer?
-            guard vm!.pointee!.pointee.GetEnv(vm, &envRaw, JNI_VERSION_1_4) == JNI_OK else {
-                fatalError("Couldn't obtain jvm environment")
+            let javaEnv = try Env.aquireJVMThread(on: vm)
+            defer {
+                try? Env.relenquishJVMThread(on: vm)
```
mstokercricut: Should be `try!`?

Sources/FishyJoesJavaRuntime/Utils.swift
```diff
@@ -65,6 +65,122 @@ public func callbackBody(
...
+    Task {
+        defer {
+            condition.lock()
+            condition.signal()
+            condition.unlock()
+        }
+
+        do {
+            let env = try Env.aquireJVMThread(on: vm)
+            print(origin.joined(separator: "\n"))
```
mstokercricut: Comment this?

Sources/FishyJoesJavaRuntime/Utils.swift
```diff
@@ -65,6 +65,122 @@ public func callbackBody(
...
+    guard let vm = try? env.GetJavaVM() else {
+        guard let errorClass = try? env.FindClass("java/lang/Error"),
+              env.ThrowNew(errorClass, "Unable to get JavaVM for environment") else {
+            fatalError("error while throwing an error")
+        }
+        return .default
+    }
+    let condition = NSCondition()
+    let resultBox: UncheckedSendableBox<Result<R, any Error>> = .init(.success(.default))
+    let origin = Thread.callStackSymbols
```
mstokercricut: Comment this?

Sources/FishyJoesJavaRuntime/Utils.swift
```diff
@@ -65,6 +65,122 @@ public func callbackBody(
...
+            } catch {
+                try! Env.relenquishJVMThread(on: vm)
+                throw error
+            }
+        } catch {
+            resultBox.value = .failure(error)
+        }
+    }
+
+    // TODO: See if there is a way to do this with Kotlin concurrency that exposes better ways to wait for values than swift...
```
mstokercricut: This may need doing before merge

integration-tests/TestAPI-bindings/kotlin/src/test/kotlin/com/cricut/testapi/AttributedStringTests.kt
```diff
@@ -1,3 +1,4 @@
+/*
```
mstokercricut: Uhh... this needs to work.

integration-tests/TestAPI-bindings/kotlin/src/test/kotlin/com/cricut/testapi/StructTests.kt
```diff
@@ -40,4 +41,16 @@ internal class StructTests {
         assertEquals(Structs.ReferenceStruct.create(), Structs.ReferenceStruct.create())
         assertNotEquals(Structs.MemberwiseStruct.create(), Structs.ReferenceStruct.create())
     }
+
+//    @Test
```
mstokercricut: This test should probably be working before we merge...

integration-tests/TestAPI/Sources/TestAPI/Functions.swift
```diff
@@ -76,4 +76,75 @@ public enum Functions {
 
     /// <!-- FishyJoes.exportReference(Functions.TheError) -->
     public struct TheError: Error {}
+
+    /// <!-- FishyJoes.export(async42Func) -->
+    public static func async42Func() async throws -> Int {
```
mstokercricut: These should be variables of function types, not methods defined on `Functions`...

integration-tests/TestAPI/Sources/TestAPI/Structs.swift
```diff
@@ -32,4 +32,26 @@ public enum Structs {
...
+        public var i = 0
+
+        /// <!-- FishyJoes.export(create) -->
+        public init() {}
+
+        public init(i: Int = 0) {
+            self.i = i
+        }
+
+//        /// <!-- FishyJoes.export(increment) -->
```
mstokercricut: These should be working before merge...

mstokercricut (changes requested):
Several things need addressing before merge.

Sources/FishyJoesJavaRuntime/Env.swift
```diff
@@ -51,6 +54,69 @@ public struct Env {
     }
 }
 
+extension Env {
+    static let jvmAttachCountKey = "fishyjoes_JVMAttachCount"
+
+    /// Calls the given body while ensuring that the body closure is attached to the JVM.
+    ///
+    /// - Important: When a suspending function is called part of the state that must be protected by calling ``Env.relenquishJVMThread(on:)`` immediately before and ``Env.aquireJVMThread(on:)`` immediately after.
+    ///     The returned `Env` should generally be assigned to the inout `javaEnv` in order to avoid shadowing and let the current environment propigate back out.
```
rodzikkonrad: "propigate" typo

Sources/FishyJoesJavaRuntime/Env.swift
rodzikkonrad: is it ok that is set to 1 in above line (when there is nothing in the dictionary under that key) and then its increased by one on the line below?

Sources/FishyJoesJavaRuntime/Env.swift
rodzikkonrad: yeah, it seems that: 
Thread.current.threadDictionary.setValue(1, forKey: Env.jvmAttachCountKey)
Should be after "guard"

Sources/FishyJoesJavaRuntime/Env.swift
```diff
@@ -121,6 +187,7 @@ extension Env {
         fns.DeleteGlobalRef(env, gref)
     }
     public func DeleteLocalRef(_ obj: jobject?) {
+//        print(Thread.callStackSymbols.joined(separator: "\n"))
```
rodzikkonrad: remove unused code

Sources/FishyJoesJavaRuntime/Env.swift
bscothern-Cricut: Fixed.

Sources/FishyJoesJavaRuntime/Env.swift
bscothern-Cricut: Fixed.

Sources/FishyJoesJavaRuntime/Env.swift
bscothern-Cricut: Yes, it defaults to 1 because something else owns the thread and we don't want to detach it. If that thread is ever detached and then calls back into swift somehow when swift attempts to acquire the thread it will detect that it is detached and set the count back to 1 so it can detach again when needed. I've added a comment helping explain the logic.

Sources/FishyJoesCore/Translators/KotlinTranslator.swift
bscothern-Cricut: Yes it can now. Fixed.

Sources/FishyJoesCore/Unparse/KotlinClass.swift
bscothern-Cricut: Fixed.

Sources/FishyJoesCore/Unparse/KotlinClass.swift
bscothern-Cricut: Fixed.

Sources/FishyJoesCore/Unparse/KotlinClass.swift
bscothern-Cricut: Ya I added the function case and never got back to updating this. Fixed.

Sources/FishyJoesCore/Unparse/KotlinClass.swift
bscothern-Cricut: It is only needed for async but doesn't hurt anything being there unless we end up building something where we don't want async symbols at all.

Sources/FishyJoesJavaRuntime/JavaFunction.swift
bscothern-Cricut: Fixed copy pasta error.

Sources/FishyJoesJavaRuntime/JavaFunction.swift
bscothern-Cricut: Fixed copy pasta error.

Sources/FishyJoesJavaRuntime/JavaFunction.swift
bscothern-Cricut: Fixed copy pasta error.

Sources/FishyJoesJavaRuntime/JavaFunction.swift
bscothern-Cricut: Fixed copy pasta error.

Sources/FishyJoesJavaRuntime/JavaFunction.swift
bscothern-Cricut: Fixed copy pasta error.

Sources/FishyJoesJavaRuntime/JavaFunction.swift
bscothern-Cricut: Fixed copy pasta error.

Sources/FishyJoesJavaRuntime/JavaReference.swift
bscothern-Cricut: Yes, otherwise special casing and logic is needed to handle primitives when being boxed to pass into/out of async land because of `@Sendable`

Sources/FishyJoesJavaRuntime/JavaReference.swift
bscothern-Cricut: Yes, that is what we decided on after I got it added in over here.

Sources/FishyJoesJavaRuntime/Utils.swift
bscothern-Cricut: Eh it was debugging stuff. I removed it.

Sources/FishyJoesJavaRuntime/Utils.swift
bscothern-Cricut: It was for debugging and has been removed.

integration-tests/TestAPI-bindings/kotlin/src/test/kotlin/com/cricut/testapi/AttributedStringTests.kt
bscothern-Cricut: Ya I forgot I turned these off...

Sources/FishyJoesJavaRuntime/Env.swift
bscothern-Cricut: Fixed.

Sources/FishyJoesJavaRuntime/Env.swift
bscothern-Cricut: Fixed.

Sources/FishyJoesCore/Translators/KotlinTranslator.swift
bscothern-Cricut: So I am reverting this back to var or let because it needs to be the force unwrapped var version for mutating functions because the reference has to escape. This might change with some of the other changes that end up happening though.

Sources/FishyJoesJavaRuntime/Env.swift.orig
```diff
@@ -0,0 +1,697 @@
...
+            Thread.current.threadDictionary[Env.jvmAttachCountKey] = jvmAttachCount + 1
+            return rawJavaEnvPointer
+                .map(OpaquePointer.init)
+                .map(UnsafeMutablePointer.init)
+                .map(Env.init(env:))!
+        case JNI_EDETACHED:
+<<<<<<< HEAD
+            Thread.current.threadDictionary[Env.jvmAttachCountKey] = 1
+=======
+>>>>>>> c39aecc (Addressed a bunch of PR comments, fixed typos, logic bugs, etc)
```
rodzikkonrad: file committed with conflicts

Sources/FishyJoesCore/Translators/KotlinTranslator.swift
bscothern-Cricut: Fixed.

Sources/FishyJoesCore/TranslatedTypes/TranslatedType.swift
bscothern-Cricut: This type is mostly needed to ensure that parameters that are functions can properly be marked as suspend funcs. I added a isSuspend value to the enum case as well so if/when it is needed it is already there since it is an important part of the function.

Sources/FishyJoesCore/Unparse/KotlinClass.swift
bscothern-Cricut: Because I am generating a bunch of Kotlin code that does dynamic and magical changing of context and the `this` object as you call the functions. The suspend func isn't just a pointer to the JNI function like our other function types. It has a body that does all the handling of that context switching and then calls out to the JNI. It is likely doable in the JNI layer but would be way way more complex to generate and manage over there.

integration-tests/TestAPI-bindings/kotlin/src/test/kotlin/com/cricut/testapi/StructTests.kt
bscothern-Cricut: Fixed.

integration-tests/TestAPI/Sources/TestAPI/Structs.swift
bscothern-Cricut: Fixed.

Sources/FishyJoesJavaRuntime/Env.swift.orig
bscothern-Cricut: Fixed.

integration-tests/TestAPI/Sources/TestAPI/Functions.swift
bscothern-Cricut: Fixed.

Sources/FishyJoesCore/Unparse/KotlinClass.swift
cobbal: Suspend callbacks and synchronous methods can now be freely mixed

Sources/FishyJoesCore/Unparse/KotlinClass.swift
cobbal: Now fixed

Sources/FishyJoesCommonRuntime/Future.swift
```diff
@@ -0,0 +1,138 @@
+import Foundation
+
+// Based on https://cs3110.github.io/textbook/chapters/ds/promises.html#the-full-implementation
+public final class Future<Output> {
```
mstokercricut, line 4: This is totally internal to FishyJoes, but I wonder if we should be adding docs to the major structure of the library as a whole. There is a lot of complexity, so documenting individual types that are used may help a future reader. The link is ultimately the best piece of documentation, but documenting in-code could help.

Sources/FishyJoesCommonRuntime/Future.swift
```diff
@@ -0,0 +1,138 @@
+import Foundation
+
+// Based on https://cs3110.github.io/textbook/chapters/ds/promises.html#the-full-implementation
```
mstokercricut, line 3: This is helpful documentation.

Sources/FishyJoesCommonRuntime/Future.swift
```diff
@@ -0,0 +1,138 @@
...
+
+    public struct Promise {
+        fileprivate let future: Future<Output>
+
+        public func resolve(_ resolved: Output) { handle(.success(resolved)) }
+        public func reject(_ error: any Error) { handle(.failure(error)) }
+        public func handle(_ result: Result<Output, any Error>) {
+            let handlers = future.withLock { () -> [Handler] in
+                if let state = future.state {
+                    fatalError("Future resolved twice, first \(state), then \(result)")
```
mstokercricut, line 20: Great check. Any fear that this actually occurs?

Sources/FishyJoesCommonRuntime/Future.swift
```diff
@@ -0,0 +1,138 @@
...
+        }
+        return promiseThunk()
+    }
+}
+
+extension Future {
+    public convenience init(_ body: @escaping () async throws -> Output) {
+        // let (self, future) = Future.make()
+        // Unfortunately, the above isn't legal swift, so `make` is inlined instead
+        self.init()
```
mstokercricut, line 108: That *is* unfortunate. Can it be `let (s, promise) = Future.make(); self = s`? Or does the compiler look for an explicit, direct call to another initializer?

Sources/FishyJoesCore/TranslatedTypes/TranslatedFunction.swift
```diff
@@ -24,9 +24,10 @@ struct TranslatedFunction: TranslatedType {
         self.sourceType = .function(parameters.map(\.sourceType), returnType.sourceType, isAsync: isAsync)
         self.neutralName = "Function<ReturnType=\(returnType.neutralName), Params=[\(parameters.map { $0.neutralName }.joined(separator: ", "))]>"
         self.nodeName = "(\(parameters.enumerated().map { "_\($0.offset): \($0.element.nodeType)" }.joined(separator: ", "))) => \(isAsync ? "Promise<" : "")\(returnType.nodeType)\(isAsync ? ">" : "")"
-        self.kotlinName = "((\(parameters.map(\.kotlinPackageQualifiedName).joined(separator: ", "))) -> \(returnType.kotlinPackageQualifiedName))"
+        let kotlinArgs = parameters.map(\.kotlinPackageQualifiedName).joined(separator: ", ")
+        self.kotlinName = "(\(isAsync ? "suspend " : "")(\(kotlinArgs)) -> \(returnType.kotlinPackageQualifiedName))"
         self.containedNamedTypes = parameters.map { $0.containedNamedTypes }.joined() + returnType.containedNamedTypes
-        self.jniType = .object("kotlin/jvm/functions/Function\(parameters.count)")
+        self.jniType = .object("kotlin/jvm/functions/Function\(parameters.count + (isAsync ? 1 : 0))")
```
mstokercricut, line 30: A bit magic. Would it be less confusing to actually add a parameter to `parameters`, or would the added complexity of doing that make it more confusing?

Sources/FishyJoesCore/Unparse/KotlinClass.swift
```diff
@@ -147,58 +145,6 @@ class KotlinClass: NestedClass {
                 if let body = method.body {
                     precondition(compatibilityParameters.isEmpty, "internal error: compatibilityOrder can't be used with a non-native body")
                     fragment.output(" = \(body)\(method.returnType.toKotlinType)")
-                } else if method.isSuspend {
```
mstokercricut: Sooooo nice.

Sources/FishyJoesJavaRuntime/Env.swift
```diff
@@ -57,63 +57,51 @@ public struct Env {
...
         guard let jvmInvoke = jvm?.pointee?.pointee else {
             throw JNIError(message: "Invalid JVM pointer")
         }
+
         // This is the only way to determine if the current thread is already attached.
         let getEnvStatus = jvmInvoke.GetEnv(jvm, &rawJavaEnvPointer, JNI_VERSION_1_4)
         switch getEnvStatus {
         case JNI_OK:
+            // If the threadDictionary doesn't have an attach count then that means something other than FishyJoes owns that thread.
+            // So by defaulting to a count of 1 we prevent detaching via FishyJoes and let the owner detach when it decides it is appropriate.
```
mstokercricut, line 71: This would be difficult to follow without the comment. Great addition.

Sources/FishyJoesJavaRuntime/Utils.swift
```diff
@@ -65,120 +70,55 @@ public func callbackBody(
...
-    let condition = NSCondition()
-    let resultBox: UncheckedSendableBox<Result<R, any Error>> = .init(.success(.default))
-    let origin = Thread.callStackSymbols
-    Task {
-        defer {
-            condition.lock()
-            condition.signal()
-            condition.unlock()
-        }
+/// Calls the given body while ensuring that the body closure is attached to the JVM.
```
mstokercricut, line 73: Good docs. I assume there is no way for us to check when this hasn't been done, right? The docs are the only way to ensure this relationship is maintained?

mstokercricut (approved):

Sources/FishyJoesCommonRuntime/Future.swift
cobbal: It's a public function, and therefore permitted by the API. Java futures have documented behavior when resolved twice, so it's very possible that that behavior will happen somewhere.

cobbal: Maybe we should pick a non-crashy behavior when it happens (just take the first result), but I'd like to know it's happening before doing that.

Sources/FishyJoesCommonRuntime/Future.swift
cobbal: With classes it doesn't work

Sources/FishyJoesCore/TranslatedTypes/TranslatedFunction.swift
cobbal: We aren't able to deal with the type on the swift side (a kotlin continuation), so it's entirely handled on the kotlin side. It does show up in the JNI interface though, which is why it's here

Sources/FishyJoesJavaRuntime/Utils.swift
cobbal: I think this one is in the "stare carefully at the code" bucket. Fortunately, not many places call this method

PR #163, opened 2024-01-20
AB#424516 Use Dart sealed class for enums and eliminate used of freezer for Dart Enums

dvosscricut:
Dart does allow for the use of named parameters in constructors, which would allow us to respect Swift named parameter names, but since the current code doesn't support that, to be backwards compatible I continued the tradition of using unnamed positional parameters in Dart, regardless of whether the Swift parameter was named or unnamed.

Freezed is still used by non enum types, namely structs defined in extensions forPuttingTypesIntoQuestionablePlaces, as well as for a struct inside the Collections enum, among others, but I thought I'd get things rolling so we can sort out particulars for this PR first before I went on that deep dive.


Sources/FishyJoesCore/Unparse/DartClass.swift
```diff
@@ -574,18 +575,25 @@ class DartEnumClass: DartClass {
...
+            } else {
+                fragment.output("\(unqualifiedName)._();")
+            }
             fragment.blankLine()
 
-            fragment.outputBlock("static int enumDiscriminator(UnownedRef obj, OutCreatedRef exn) => check((exn) =>", closeWith: ");") {
+            fragment.outputBlock("static int enumDiscriminator(UnownedRef obj, OutCreatedRef exn) => check((exn) {", closeWith: "});") {
                 if cases.isEmpty {
-                    fragment.output("throw UnsupportedError('This class is supposed to be unihabited')")
+                    fragment.output("throw UnsupportedError('This class is supposed to be unihabited');")
```
mstokercricut: `uninhabited`

Sources/FishyJoesCore/Unparse/DartClass.swift
```diff
@@ -574,18 +575,25 @@ class DartEnumClass: DartClass {
...
-                        for (index, enumCase) in cases.enumerated() {
-                            fragment.output("\(enumCase.name): (_) => \(index),")
+                    fragment.output("final peekedObj = peekRef<\(unqualifiedName)>(obj);")
+                    for (enumIndex, enumCase) in cases.enumerated() {
+                        if enumIndex != 0 {
+                            fragment.output("else ", newLineTerminated: false)
                         }
+                        fragment.output("if (peekedObj is \(unqualifiedName)_\(upperCaseFirst(enumCase.name))) { return \(enumIndex); }")
                     }
+                    fragment.output("throw UnsupportedError('Unknown \(unqualifiedName) subclass');")
```
mstokercricut: The previous implementation did not throw from what I could tell. Maybe the `map` could somehow. Can you see any way we can eliminate the need to throw?

Sources/FishyJoesCore/Unparse/DartClass.swift
```diff
@@ -574,18 +575,25 @@ class DartEnumClass: DartClass {
...
                 if cases.isEmpty {
-                    fragment.output("throw UnsupportedError('This class is supposed to be unihabited')")
+                    fragment.output("throw UnsupportedError('This class is supposed to be unihabited');")
                 } else {
-                    fragment.outputBlock("peekRef<\(unqualifiedName)>(obj).map(", closeWith: ")") {
-                        for (index, enumCase) in cases.enumerated() {
-                            fragment.output("\(enumCase.name): (_) => \(index),")
+                    fragment.output("final peekedObj = peekRef<\(unqualifiedName)>(obj);")
+                    for (enumIndex, enumCase) in cases.enumerated() {
+                        if enumIndex != 0 {
```
mstokercricut: Since these all return, there is no need for `else if`. If you do want to keep the `else if` structure, though, the final statement should be an `else`.

Sources/FishyJoesCore/Unparse/DartClass.swift
```diff
@@ -643,6 +651,78 @@ class DartEnumClass: DartClass {
...
+                fragment.blankLine()
+
+                for value in enumCase.values {
+                    fragment.output("final \(value.type.name(in: self)) \(DartClass.deforbidify(value.name));")
+                }
+
+                fragment.blankLine()
+
+                fragment.output("@override")
+                fragment.output("bool operator ==(Object other)", newLineTerminated: false)
```
mstokercricut, line 685: I didn't see any tests for `==` or `hashCode`. Since we're now taking charge of those operations, we should test them in `TestAPI-bindings` for Dart.

E.g. test if self equality works, if two instances of the same case work, test a few associated data enums for equality, and test if equal instances have the same hash value.

A test placing all of the cases of an enum into a Set would probably also be called for, ensuring that they can be used correctly as option flag collections.

Sources/FishyJoesCore/Unparse/DartClass.swift
```diff
@@ -550,21 +550,22 @@ class DartEnumClass: DartClass {
     }
 
     override var hasFreezedPart: Bool {
```
mstokercricut, line 552: References to Freezed should eventually be completely eliminated.

mstokercricut (changes requested):
This is an excellent start! If you'd like to merge into main after addressing these comments, that would be fine. However, there is little need to do a versioned release before completely eliminating Freezed.

Sources/FishyJoesCore/Unparse/DartClass.swift
dvosscricut: we could have an else that returned 0 (the first enum case).

However I believe throwing and exception is quite appropriate if we ever get an object that is not one of the known enum cases.

Sources/FishyJoesCore/Unparse/DartClass.swift
dvosscricut: yes, I could do an else as the final statement and throw the exception there.

Sources/FishyJoesCore/Unparse/DartClass.swift
dvosscricut: The tests exist in the enumTests. Any expect(..., equals()) uses it. (enum_test.dart).

If you don't have the == and hashCode, then the tests in enum_test.dart won't work

Sources/FishyJoesCore/Unparse/DartClass.swift
dvosscricut: I agree!

I'm just doing it for enums right now, and structs still use it, so this computed var must still exist.

Sources/FishyJoesCore/Unparse/DartClass.swift
dvosscricut: good catch

Sources/FishyJoesCore/Unparse/DartClass.swift
dvosscricut: it's just more complicated to generate that equivalently functioning code. But I will do it, just for you, and for style.

dvosscricut: actually it wasn't too bad to add it. done!

Sources/FishyJoesCore/Unparse/DartClass.swift
dvosscricut: I decided to go with the if/else branching structure instead of the map because it's the same amount of code in the enum discriminator but it cuts down on a lot of bloated boilerplate 'map' code in each the sealed class's subclasses.

It shouldn't be possible to actually reach this else case, but if it is an exception is the proper way to warn/deal with it.

map used overridden subclass functions to pick the appropriate closure passed into it from the many closures passed in (one per subclass). But if a subclass for some reason did not override the map function you would get an exception because it would call the parent map function which does a `throw _privateConstructorUsedError `because it's supposed to be always overridden by a subclass. So it's kind of the same dealy-oh here, just in a different form. If the generator forgot to put in an else case for a subclass, it would throw an unknown subclass type exception in this PR code. On the freezer map type code, it would throw the _privateConstructorUsedError if the generator somehow forgot to override the map function in a subclass.

Sources/FishyJoesCore/Unparse/DartClass.swift
dvosscricut: regardless, I'll make some explicit tests for it.

dvosscricut:
> This is an excellent start! If you'd like to merge into main after addressing these comments, that would be fine. However, there is little need to do a versioned release before completely eliminating Freezed.

@mstokercricut  
I am blocked from merging without an approval.

Sources/FishyJoesCore/Unparse/DartClass.swift
mstokercricut: 👍🏿

Sources/FishyJoesCore/Unparse/DartClass.swift
mstokercricut: I was more concerned that the relationship between `equals` and `hashCode` was being respected, and a test for that was made part of the test suite.

mstokercricut (approved):
This is a pure improvement to existing code, so it can go into main. Hold off on a released version until Freezed is eliminated, though.

PR #164, opened 2024-01-22
AB#426027 Put in toString, shallowCopy, and test for hashCode on dart sealed class enums

dvosscricut:
AB#426027 Put in toString, shallowCopy, and test for hashCode on dart sealed class enums

Dart doesn't support any kind of built in deepCopy so I went with a shallowCopy: https://stackoverflow.com/questions/13107906/how-can-i-clone-an-object-deep-copy-in-dart/67420251#67420251

### DIFF
--- a/Sources/FishyJoesCore/Unparse/DartClass.swift
+++ b/Sources/FishyJoesCore/Unparse/DartClass.swift
@@ -699,9 +699,11 @@ class DartEnumClass: DartClass {
                         fragment.blankLine()
                     } else {
                         fragment.output(" &&")
-                        fragment.outputMap(enumCase.values, separator: " &&") { value in
-                            let valueName = "\(DartClass.deforbidify(value.name))"
-                            return "(identical(other.\(valueName), \(valueName)) || other.\(valueName) == \(valueName))"
+                        fragment.outputBlock("(") {
+                            fragment.outputMap(enumCase.values, separator: " &&") { value in
+                                let valueName = "\(DartClass.deforbidify(value.name))"
+                                return "const DeepCollectionEquality().equals(other.\(valueName), \(valueName))"
+                            }
                         }
                     }
                     fragment.currentIndent -= 1
@@ -720,7 +722,7 @@ class DartEnumClass: DartClass {
                     fragment.currentIndent += 1
                     fragment.output("runtimeType,")
                     fragment.outputMap(enumCase.values, separator: ", ") { value in
-                        "\(DartClass.deforbidify(value.name))"
+                        "const DeepCollectionEquality().hash(\(DartClass.deforbidify(value.name)))"
                     }
                     fragment.currentIndent -= 1
                     fragment.output(");")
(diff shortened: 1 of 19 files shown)


Sources/FishyJoesCore/Unparse/DartClass.swift
```diff
@@ -650,7 +650,15 @@ class DartEnumClass: DartClass {
             methods.forEach { output(method: $0, to: fragment) }
 
             fragment.blankLine()
+
             outputNativeMethodDeclarations(to: fragment)
+
+            fragment.blankLine()
+
+            fragment.output("\(unqualifiedName) shallowCopy()", newLineTerminated: false)
+            fragment.outputBlock(" {") {
```
cobbal: nit, more idiomatic in this codebase to combine the 2 lines:
```suggestion
            fragment.outputBlock("\(unqualifiedName) shallowCopy() {") {
```

Sources/FishyJoesCore/Unparse/DartClass.swift
```diff
@@ -720,6 +728,40 @@ class DartEnumClass: DartClass {
...
+                fragment.outputBlock(" {") {
+                    fragment.output("return \(unqualifiedName).\(enumCase.name)(", newLineTerminated: false)
+                    var paramsString = String()
+                    for (index, value) in enumCase.values.enumerated() {
+                        let valueString = "\(DartClass.deforbidify(value.name))"
+                        paramsString += "\(valueString)"
+                        if index < enumCase.values.indices.upperBound - 1 {
+                            paramsString += ", "
+                        }
+                    }
```
cobbal, line 747: nit: `.map(...).joined(", ")` would be a cleaner way to get paramString in both these cases

cobbal (approved):
Approved with nits

integration-tests/TestAPI-bindings/dart/test/enum_test.dart
```diff
@@ -1,3 +1,5 @@
+import 'dart:js_util';
```
cobbal: This line is causing CI to fail. It looks related to dart-web? which we don't target

Sources/FishyJoesCore/Unparse/DartClass.swift
dvosscricut: outputMap puts it on multiple lines. I wanted it on a single line.

integration-tests/TestAPI-bindings/dart/test/enum_test.dart
dvosscricut: why does it keep adding that? I'll remove it again.

Sources/FishyJoesCore/Unparse/DartClass.swift
cobbal: The whole loop could be replaced by `let paramsString = enumCase.values.map { DartClass.deforbidify($0.name) }.joined(separator: ", ")` and the value of paramString would be exactly the same

mstokercricut (approved):

Sources/FishyJoesCore/Unparse/DartClass.swift
dvosscricut: ok

Sources/FishyJoesCore/Unparse/DartClass.swift
dvosscricut: good point

PR #165, opened 2024-01-24
AB#424516 Replace freezed gnr8d code for structs with public constructors with our own gnr8d code

dvosscricut:
AB#424516 Replace freezed gnr8d code for structs with public constructors with our own gnr8d code

I discovered that for the Set, List, and Map there is a built in copy method Set.from(someSet) that I used in shallowCopy.

I also discovered that there is a 20 argument limit for positional parameters input to Object.hash(), so I had to write some code to dive it up into sub Object.hashes input to a top level Object.hash to handle Primitives_PrimitiveHolder. So there is a limit of 20*20=400 positional parameters in structs with public constructors now. That should be enough ("640kiB should be enough for everybody" <:o) ).

I also discovered that DeepCollectionEquality().equals is needed to test if the Set/List/Map types are equal. I found they worked as well for primitive types, so now we just use DeepCollectionEquality().equals in the == operator function instead of what we had before.


Sources/FishyJoesCore/Unparse/DartClass.swift
```diff
@@ -740,12 +816,19 @@ class DartEnumClass: DartClass {
                 fragment.output("@override")
                 fragment.output("\(unqualifiedName) shallowCopy() => \(unqualifiedName).\(enumCase.name)", newLineTerminated: false)
                 if !enumCase.values.isEmpty {
-                    fragment.outputBlock(" (", newLineTerminated: false) {
+                    fragment.outputBlock(" (", closeWith: ");") {
                         fragment.outputMap(enumCase.values, separator: ", ") {
-                            "\(DartClass.deforbidify($0.name))"
+                            switch $0.type {
+                            case let .named(_, name, _):
+                                if name == "Set" || name == "List" || name == "Map" {
```
mstokercricut: This seems like a brittle way to identify this case. What if I want a `Cartography.Map`?

integration-tests/TestAPI-bindings/dart/lib/src/generated/Primitives_PrimitiveHolder.dart
```diff
@@ -543,6 +566,111 @@ class Primitives_PrimitiveHolder with _$Primitives_PrimitiveHolder {
...
+                const DeepCollectionEquality().equals(other.d, d) &&
+                const DeepCollectionEquality().equals(other.dq, dq)
+            )
+        );
+    }
+
+    @override
+    int get hashCode => Object.hash(
+        runtimeType,
+        Object.hash(
```
mstokercricut, line 586: Good catch on this! Glad you found a way to make it work.

Sources/FishyJoesCore/Unparse/DartClass.swift
```diff
@@ -433,16 +425,19 @@ class DartProductClass: DartClass {
...
-                fragment.outputBlock("factory \(unqualifiedName)({", closeWith: "})", newLineTerminated: false) {
+                for field in fields {
+                    fragment.output("\(field.isPubliclyWritable ? "" : "final ")\(field.type.name(in: self)) \(DartClass.deforbidify(field.name));")
+                }
+
+                fragment.blankLine()
+
+                fragment.outputBlock("\(unqualifiedName)({", closeWith: "});") {
                     fragment.outputMap(fields, separator: ",") { field in
-                        "required \(field.isPubliclyWritable ? "" : "final ")\(field.type.name(in: self)) \(DartClass.deforbidify(field.name))"
```
mstokercricut, line 438: Allowing a field that is not publicly writable on the Swift side to be modified causes the interfaces to not match in terms of mutability. Could you expand on why this change is necessary? I saw the test below with `a.x`, but didn't follow the logic...

mstokercricut (changes requested):

Sources/FishyJoesCore/Unparse/DartClass.swift
dvosscricut: The 'final' is now in the property declaration list rather than the factory constructor because the factory constructor has been removed in favor of the generative constructor. It's still there though.

Sources/FishyJoesCore/Unparse/DartClass.swift
dvosscricut: As I understand it, the 'name' is really a type (since it's under the DartType type), so Cartography.Map would need to be named Cartography_Map instead because Map/Set/List are built in Dart types. There shouldn't be any naming conflicts

dvosscricut: additionally, if somehow there could be a type named "Cartography.Map", the string compare would fail here because it's != "Map" because it compares the entire string.

mstokercricut (approved):

Sources/FishyJoesCore/FishyJoesContext.swift
```diff
@@ -124,7 +124,6 @@ public class FishyJoesContext {
             } + dartClasses.flatMap { cls in
                 [
                     "import './\(cls.unqualifiedName).dart' as \(module);",
-                    "import './\(cls.unqualifiedName).dart';", // Import unqualified too or freezed gets confused"
```
cobbal, line 127: Glad this can be removed

Sources/FishyJoesCore/Unparse/DartClass.swift
```diff
@@ -507,6 +502,96 @@ class DartProductClass: DartClass {
...
+                fragment.output("\(toStringParamsString))';")
+
+                fragment.blankLine()
+
+                fragment.output("@override")
+                fragment.output("bool operator ==(Object other)", newLineTerminated: false)
+                fragment.outputBlock(" {") {
+                    fragment.output("return identical(other, this) ||")
+                    fragment.output("(")
+                    fragment.currentIndent += 1
```
cobbal: "outputBlock" is the best way to manage indent bookkeeping

Sources/FishyJoesCore/Unparse/DartClass.swift
```diff
@@ -507,6 +502,96 @@ class DartProductClass: DartClass {
...
+                            }
+                            fragment.blankLine()
+                        }
+                    }
+                }
+
+                fragment.blankLine()
+
+                fragment.output("@override")
+                fragment.output("\(unqualifiedName) shallowCopy() => \(unqualifiedName)", newLineTerminated: false)
```
cobbal: I can't find the method this is overriding, and I don't see a "shallowCopy" in the old freezed files. Do you have a sample you were basing this on?

cobbal (changes requested):

Sources/FishyJoesCore/Unparse/DartClass.swift
dvosscricut: good point

Sources/FishyJoesCore/Unparse/DartClass.swift
dvosscricut: you're right there is no override for the DartProductClass. Good catch.

mstokercricut (approved):
LGTM

cobbal (approved):

PR #167, opened 2024-01-31
AB#427203 Translate Swift Protocols to Dart

dvosscricut:
AB#427203 Translate Swift Protocols to Dart

Draft because probably should implement other languages C#, Wasm, and Kotlin before a merge so we don't merge strange code.


Sources/FishyJoesCore/FishyJoesContext.swift
```diff
@@ -152,10 +152,16 @@ public class FishyJoesContext {
 
         // Collect type information before starting translation
         // This collects the named types possible for use later in resolve().
-        let translatedTypes = templateContext.types.all.compactMap { type -> TranslatedType? in
+        var translatedTypes = templateContext.types.protocols.compactMap { type -> TranslatedType? in
```
cobbal: In #29 this was `types.types` to get both types and protocols

Sources/FishyJoesCore/ExportAnnotation.swift
```diff
@@ -19,6 +20,7 @@ struct ExportAnnotation: Hashable {
     let genericOverrides: [String: BetterType]
     let omitParameters: [String]
     let compatibilityOrder: [String]
+    let protocols: [String]
```
cobbal, line 23: We should be able to detect which conformances are exported without manual annotation. See the "matchedConformances" line in #29

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
```diff
@@ -0,0 +1,253 @@
...
+        self.sourceType = BetterType(named: type, context: context)
+        self.nodeName = name
+        self.kotlinName = name
+        self.neutralName = "Protocol<TranslatedFrom=\(name)>"
+        self.kotlinPackage = context.module.kotlinPackage
+        self.cSharpType = .named(package: context.module.cSharpNamespace, name: exportAnnotation.cSharpName)
+        self.dartType = .named(package: context.module.dartNamespace, name: context.dartTranslator.fakeNamespace(name))
+        self.jniType = .object(context.kotlinTranslator.javaClassName(nodeName, in: context))
+
+        self.storedVariables = type.storedVariables
```
cobbal, line 34: I didn't think a protocol could have stored variables

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
```diff
@@ -0,0 +1,253 @@
...
+        self.computedVariables =
+            (type.computedVariables + type.staticVariables).filter { $0.exportAnnotation != nil }
```
cobbal, line 36: There's an inconsistency here between the methods (which don't require an export annotation) and the variables (which do).

integration-tests/TestAPI/Sources/TestAPI/Protocols.swift
```diff
@@ -0,0 +1,179 @@
...
+        } else {
+            return 42.909
+        }
+    }
+
+    /// <!-- FishyJoes.export(spqr) -->
+    public func spqr(_ pippo: AssociatedDataEnum) -> Int {
+        pippo.intValue
+    }
+}
```
cobbal, line 179: There are 2 more things I would expect to see in protocol tests. First is protocols being passed from swift to dart by their type (e.g. `var aProtocolObject: TestMethodsProtocol`) and second is them being passed back (e.g. `func useProtocolObject(aProtocolObject: TestMethodsProtocol)`

Sources/FishyJoesCore/FishyJoesContext.swift
```diff
@@ -542,6 +550,8 @@ public class FishyJoesContext {
 
     func add(dartClass: DartClass) {
         dartClasses.append(dartClass)
+        // Don't add to native methods for DartProtocolClass, since there is nothing to export for a protocol.
+        if dartClass is DartProtocolClass { return }
```
cobbal, line 548: DartProtocolClass should just have an empty "nativeMethods" then, instead of being dynamically type-sniffed

Sources/FishyJoesCore/Method.swift
```diff
@@ -69,16 +73,24 @@ struct Method: Hashable {
...
         var omitParameters = Set(exportAnnotation.omitParameters)
+        var placeholderNum = 1
         for parameter in method.parameters {
             if omitParameters.contains(parameter.name) {
                 precondition(parameter.defaultValue != nil, "Can't omit non-default parameter")
                 omitParameters.remove(parameter.name)
                 continue
             }
+            var name = parameter.name
+            // If param is for protocol method, we need to put in a placeholder name if the swift param is unnamed, because languages like Dart insist on parameter names for their abstract class methods.
```
cobbal, line 84: This seems odd to me. Do you have an example of a dart method signature that is legal in a concrete class, but disallowed in an abstract one?

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
cobbal: I'm not 100% sure which is right.

"just not exporting a method" doesn't really work for a required method of a protocol, since there would be no way for a dart type to then conform to the protocol.

I do like being able to name the exported methods more idiomatically though, so export annotations are good

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
```diff
@@ -0,0 +1,253 @@
...
+        return [fragment]
+    }
+
+    func nodeDefinitionFragment(in context: FishyJoesContext) -> SourceFragment {
+        let fragment = context.swiftFragment(
+            "NodeInterface/\(sourceType.name)+node.swift",
+            additionalImports: ["Foundation", "FishyJoesNodeRuntime"]
+        )
+        fragment.blankLine()
+        return fragment
```
cobbal, line 89: empty fragments are probably what is causing the swiftlint errors. Just delete this method for now.

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
```diff
@@ -0,0 +1,253 @@
...
+                            fragment.output(resolved.cSharpType.isObject ? "newValue.Consume<\(resolved.cSharpType.name)>();" : "newValue;")
+                        }
+                    },
+                ]
+            )
+        }
+    }
+
+    func dartSetupDelegates(in context: FishyJoesContext) -> [String] {
+        return []
```
cobbal, line 182: I'm suprised there's no runtime communication between the languages for a protocol. This will probably change once support for passing protocols back and forth is added.

Sources/FishyJoesCore/Unparse/DartClass.swift
```diff
@@ -835,6 +866,108 @@ class DartEnumClass: DartClass {
     }
 }
 
+class DartProtocolClass: DartClass {
+    struct Typealias {
```
cobbal: I don't think Typealias is used

Sources/FishyJoesCore/Unparse/DartClass.swift
```diff
@@ -835,6 +866,108 @@ class DartEnumClass: DartClass {
...
+        let name: String
+        let value: DartType
+    }
+
+    enum Constructor: Equatable {
+        case `public`(fields: [Variable])
+        case reference
+    }
+
+    let constructor: Constructor
```
cobbal: constructor is written, but never read. I don't think it makes sense for this abstract class to have a constructor anyway.

Sources/FishyJoesCore/Unparse/DartClass.swift
```diff
@@ -835,6 +866,108 @@ class DartEnumClass: DartClass {
...
+        commonIgnoreSpecificWarnings(fragment: fragment)
+
+        document(documentation, fragment: fragment)
+
+        fragment.output("abstract class \(unqualifiedName)", newLineTerminated: false)
+        fragment.outputBlock(" {", closeWith: "}") {
+            for method in methods {
+                // Dart does not support static method inheritance like Swift does. So we will put it in the subclass, but not in the abstract class that the subclass implements, for that would result in a compile error.
+                guard !method.isStatic else {
+                    fragment.output("// \(method.name) not declared in abstract class because Dart does not support static method inheritance. It will only be put into the subclass that implements this abstract class.")
```
cobbal, line 895: This could be a problem... should we just forbid exporting protocols with required static methods?

integration-tests/TestAPI-bindings/dart/lib/src/generated/TestMethodsProtocol.dart
```diff
@@ -0,0 +1,69 @@
...
+
+// ignore_for_file: unused_import
+// ignore_for_file: non_constant_identifier_names
+// ignore_for_file: no_leading_underscores_for_local_identifiers
+// ignore_for_file: library_prefixes
+// ignore_for_file: file_names
+// ignore_for_file: annotate_overrides
+
+/// <!-- FishyJoes.export(TestMethodsProtocol) -->
+abstract class TestMethodsProtocol {
```
cobbal, line 51: I don't fully understand the difference between "abstract class" and "interface class" in dart, but was suprised to learn you can implement a class you don't inherit from.

cobbal (changes requested):
This looks like a great start. The biggest thing that I see missing is being able to send and receive protocol objects.

Sources/FishyJoesCore/ExportAnnotation.swift
dvosscricut: we can detect them, yes, but I don't necessarily want to implement all protocols a thing conforms to in swift in the target language. For example, Hashable. I don't want to export that to Dart because Swift handles that already. It's not a user implemented protocol.

We only want to export user implemented protocols (I think).

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
dvosscricut: In swift a protocol can, it's a `var foo: Int { set get }`

integration-tests/TestAPI/Sources/TestAPI/Protocols.swift
dvosscricut: ok

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
dvosscricut: All I did for protocols in Dart was use an abstract class with method and property (set and/or get) definitions. Then the objects that use a protocol do an 'implements' of it.

I didn't seem to need any runtime anything for that, it's just declaring an interface. 

Since the actual object that implements the abstract class is where the runtime back and forth is.

Sources/FishyJoesCore/Unparse/DartClass.swift
dvosscricut: perhaps not. This was a copy and paste job. Can certainly be cleaned up.

Sources/FishyJoesCore/Unparse/DartClass.swift
dvosscricut: well, we can support it. We just don't put it in the abstract base class, we put it in the class that implements that abstract base class.

But you couldn't do "AbstractBaseClassName.staticMethodName()" in Dart. You'd have to do "ClassThatImplementsAbstractBaseClassName.staticMethodName()". 

If that's not useful we can prohibit static method and property exports for protocols.

integration-tests/TestAPI-bindings/dart/lib/src/generated/TestMethodsProtocol.dart
dvosscricut: implements is like a protocol, you have to implement ALL the properties and methods in the abstract base class.

extends is the inheritance. You can extend an abstract base class too. But if you extend a normal class you get the inheritance of the non overridden methods/properties.

You can only extend from one class i.e. inherit from one class.

But you can 'implement' many classes. So 'implements' with abstract base classes seems the way to go for Swift protocol definitions in Dart.

Sources/FishyJoesCore/Unparse/DartClass.swift
dvosscricut: good point. removed.

Sources/FishyJoesCore/Unparse/DartClass.swift
dvosscricut: good point. removed.

Sources/FishyJoesCore/FishyJoesContext.swift
dvosscricut: good point. fixed

Sources/FishyJoesCore/ExportAnnotation.swift
dvosscricut: I noticed in your `cobbal/protocols` branch that you have a hard coded lists of protocols to exclude, including Hashable. However I believe it's more flexible to have the user declare what protocols they want exported, because the list of hard coded protocols may miss something to exclude. It's also easier to implement and maintain.

cobbal: Yeah. On second read that line isn't doing quite what I thought it was doing. I'm still on the fence if it's something we need to force the user to deal with. Seems like something we should be able to figure out for them.

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
cobbal: I believe that's a computed variable (at least in the protocol)

dvosscricut: For example, in the TestOptionalsProtocol, `flarp` shows up in the storedVariables in TranslatedProtocol.swift initializer.

In the TestProtocolClass in Protocols.swift, flare is implemented as a property, not a computed variable.

/// 
    public var flarp: String?

Sources/FishyJoesCore/ExportAnnotation.swift
dvosscricut: It's more error prone to try and figure out for the user. We're already having the user export objects, methods, and properties, why not protocols as well?

That way, if the user's object conforms to some protocol we've never heard of that's an Apple Protocol, FishyJoes won't mess up on it saying it can't find it, because they won't export it. They'll only export the protocols that they define (and want to export).

dvosscricut:
archive for future reference

PR #173, opened 2024-02-20
AB#427207 Simple Protocol Support Export to Kotlin

dvosscricut:
AB#427207 Simple Protocol Support Export to Kotlin

Includes:

Default Implementations in Protocols

Excludes:

Static var/func support in Protocols

Protocol conformances to export to Fishy Joes are designated in export annotation in order to simplify implementation and allow user control of what Protocols are exported. The implementation is simplified because then FishyJoes doesn't have to somehow auto detect what Protocols to generate code for and what not to, which could be complicated and error prone.

Please find tests in `integration-tests/TestAPI-bindings/kotlin/src/test/kotlin/com/cricut/testapi/ProtocolTests.kt`


Sources/FishyJoesCore/Method.swift
```diff
@@ -38,7 +39,8 @@ struct Method: Hashable {
         isMutating: Bool = false,
         isThrowing: Bool = false,
         isAsync: Bool = false,
-        deprecation: Deprecation? = nil
+        deprecation: Deprecation? = nil,
+        implemented: Bool = false
```
cobbal: I would expect this to default to true. Aren't all methods for non-protocol types implemented? Maybe I don't understand what this boolean means. Either renaming, changing the default, or documentation could help, but I can't tell which.

Sources/FishyJoesCore/Method.swift
```diff
@@ -82,14 +86,23 @@ struct Method: Hashable {
             parameters.append(
                 SwiftFormal(
                     label: parameter.argumentLabel,
-                    name: parameter.name,
+                    name: isProtocolDef ? (parameter.asSource.starts(with: "_ ") ? "" : parameter.name) : parameter.name,
```
cobbal: This feels a little odd having an empty string as the name. There might be a cleaner place to deal with whatever problem "name" was causing.

Sources/FishyJoesCore/Method.swift
```diff
@@ -100,3 +113,43 @@ extension Method: CustomStringConvertible {
         return "\(namespace)\(callName)(\(params))"
     }
 }
+
+extension Method {
+    static func javaClassName(_ name: String, in context: FishyJoesContext) -> String {
+        "com/cricut/\(context.module.name.lowercased())/\(name.replacingOccurrences(of: ".", with: "$"))"
+    }
+
+    func jniSignature(context: FishyJoesContext) -> String {
```
cobbal, line 124: Probably a good extension method to have, but I think it would make more sense in KotlinTranslator.swift

Sources/FishyJoesCore/Method.swift
```diff
@@ -100,3 +113,43 @@ extension Method: CustomStringConvertible {
...
+        }
+        let returnType = context.resolve(type: returnType, generics: exportAnnotation.genericOverrides)
+        jniSignature = "(\(jniSignature))\(isAsync ? "Lkotlinx/coroutines/Deferred;" : returnType.jniType.asSignature)"
+        return jniSignature
+    }
+}
+
+extension Method {
+    func swiftClosureSignature() -> String {
+        "(\(parameters.map { $0.type.name}.joined(separator: ", "))) -> \(returnType.name)"
```
cobbal: This will be missing "async" and "throws" annotations.

Sources/FishyJoesCore/TranslatedTypes/TranslatedType.swift
```diff
@@ -14,6 +14,7 @@ protocol TranslatedType {
     var definingTSNamespace: String? { get }
     var isInhabited: Bool { get }
     var containedNamedTypes: [TranslatedType] { get }
+    var conformances: Set<String> { get }
```
cobbal, line 17: "String" may need to be changed in the future to be a full BetterType. In case some maniac decides to name protocols in 2 libraries the same thing. Then we'd need to be able to work with a (sometimes) package qualified name.

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
```diff
@@ -0,0 +1,349 @@
...
+    let definingModule: Module
+    let definingTSNamespace: String?
+    let isInhabited: Bool
+    var containedNamedTypes: [TranslatedType] { [self] }
+    let conformances: Set<String>
+    let methods: [Method]
+    let computedVariables: [Variable]
+    let documentation: [String]
+    let className: String
+    let externalWitnessClassName: String
```
cobbal: This field seems java-specific. If so, the name should reflect that (or it should just live entirely in the java-specific files)

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
```diff
@@ -0,0 +1,349 @@
...
+    func jniDefinitionFragments(in context: FishyJoesContext) -> [SourceFragment] {
+        let fragment = context.swiftFragment(
+            "JavaInterface/\(sourceType.nonNamespacedName)+java.swift",
+            additionalImports: ["Foundation", "FishyJoesJavaRuntime", "CommonInterface"]
+        )
+
+        let foreignProtocolType = "_Java\(sourceType.nonNamespacedName)"
+
+        var methodIDs: [(idHandle: String, name: String, signature: String)] = []
+        let defaultMethods = methods.filter { $0.implemented }
```
cobbal: would "isDefaultImplementation" be a more accurate name for "implemented"?

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
```diff
@@ -0,0 +1,349 @@
...
+        let normalMethods = methods.filter { !$0.implemented }
+
+        fragment.outputBlock("struct \(foreignProtocolType): \(sourceType.nonNamespacedName) {") {
+            fragment.output("let _javaWitness: JavaReference")
+            let hasStaticMethods = methods.contains(where: { $0.isStatic })
+            let hasStaticVariables = computedVariables.contains(where: { $0.isStatic })
+            if hasStaticMethods || hasStaticVariables {
+                fragment.output("static let _javaWitness: JavaReference = { _javaWitness }()")
```
cobbal: Is this an infinite loop?

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
```diff
@@ -0,0 +1,349 @@
...
+                let getID = "_\(name)GetMethodID"
+                let setID = variable.isMutable ? "_\(name)SetMethodID" : nil
+                methodIDs.append((idHandle: getID, name: "get_\(name)", signature: "()\(resolved.jniType.asSignature)"))
+                fragment.output("static var \(getID): jmethodID?")
+                if let setID = setID {
+                    methodIDs.append((idHandle: setID, name: "set_\(name)", signature: "(\(resolved.jniType.asSignature))V"))
+                    fragment.output("static var \(setID): jmethodID?")
+                }
+                fragment.outputBlock("\(variable.isStatic ? "static " : "")public var \(name): \(type) {") {
+                    func gnr8GetterCode(fragment: SourceFragment) {
```
cobbal: nope :)

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
```diff
@@ -0,0 +1,349 @@
...
+                        fragment.output("let env = try! _javaWitness.currentThreadEnv()")
+                        fragment.outputBlock("return try! \(resolved.converterType.name).fromJava(") {
```
cobbal: Hmm... we currently refuse to translate non-throwing function types because kotlin can't express them. Having all getters on all protocols need to be throwing might be a bigger user headache though. Is allowing them and then crashing at runtime if an exception happens better or worse long-term? @mstokercricut?

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
```diff
@@ -0,0 +1,349 @@
...
+                    }
+                }
+            }
+            for method in methods {
+                fragment.output()
+                let resolvedReturn = context.resolve(type: method.returnType)
+                var returnSignature = "\(method.isThrowing ? " throws" : "")"
+                if method.returnType.name != "Void" {
+                    returnSignature.append(" -> \(method.returnType.name)")
+                }
```
cobbal, line 316: This is the logic that should also go in Method.swiftClosureSignature

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
```diff
@@ -0,0 +1,349 @@
...
+                fragment.output("static var _\(method.callName)MethodID: jmethodID?")
+                if method.implemented {
+                    fragment.output("\(method.isStatic ? "static " : "")public var \(method.callName)Impl: (\(method.swiftClosureSignature()))?")
+                }
+                var paramSigs = [String]()
+                do {
+                    var unnamedParamCnt = 1
+                    for param in method.parameters {
+                        var paramName = param.name
+                        if paramName.isEmpty {
```
cobbal: This is where the empty name mentioned before comes back. I'm still not seeing the reason the name needed to be replaced with empty though. Maybe it comes up again later.

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
```diff
@@ -0,0 +1,349 @@
...
+            additionalImports: ["Foundation", "FishyJoesJavaRuntime", "CommonInterface"]
+        )
+
+        let foreignProtocolType = "_Java\(sourceType.nonNamespacedName)"
+
+        var methodIDs: [(idHandle: String, name: String, signature: String)] = []
+        let defaultMethods = methods.filter { $0.implemented }
+        let normalMethods = methods.filter { !$0.implemented }
+
+        fragment.outputBlock("struct \(foreignProtocolType): \(sourceType.nonNamespacedName) {") {
```
cobbal: nonNamespacedName should be used as sparingly as possible. The namespaced version should be preferred everywhere it doesn't break something.

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
```diff
@@ -0,0 +1,349 @@
...
+                }
+                fragment.outputBlock("\(method.isStatic ? "static " : "")public func \(method.callName)(\(paramSigs.joined(separator: ", ")))\(returnSignature) {") {
+                    if method.implemented {
+                        fragment.output("guard let \(method.callName)Impl = \(method.callName)Impl else", newLineTerminated: false)
+                        fragment.outputBlock(" {", closeWith: "}") {
+                            let methodCallStr = "\(method.callName)(\(method.parameters.map { "\($0.name): \($0.name)" }.joined(separator: ", ")))"
+                            fragment.output("return \(foreignProtocolType)_sans_\(method.callName)(wrapped: self).\(methodCallStr)")
+                        }
+                        fragment.output("return \(method.callName)Impl(\(method.parameters.map { $0.name }.joined(separator: ", ")))")
+                    } else {
```
cobbal: The organization surprises me a little here. I would expect the "else" of this block to look similar to the "if" block, only differing in the default fallback. One, however, uses a stored "Impl" closure property while the other one uses the JNI's call...Method. I suspect this means something could be cleaned up.

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
```diff
@@ -0,0 +1,349 @@
...
+                    fragment.output("return try Box<SwiftType>.fromJava(value, env: env).value")
+                }
+                fragment.output("return \(foreignProtocolType)(_javaWitness: try JavaReference(local: value, env: env))")
+            }
+
+            fragment.outputBlock("public static func toJava(_ value: SwiftType, env: Env) throws -> jobject? {") {
+                fragment.outputBlock("try env.NewObject(") {
+                    fragment.output("externalWitnessClass,")
+                    fragment.output("externalWitnessConstructor,")
+                    fragment.output("jvalue(j: jlong(UInt(bitPattern: Box(value).retainedOpaque())))")
```
cobbal: I think there was a convenience constructor added `jvalue(pointer:)` to make this less painful

Sources/FishyJoesCore/TranslatedTypes/TranslatedType.swift
```diff
@@ -210,3 +213,63 @@ extension Type {
...
+        isClass == other.isClass &&
+        isFailableInitializer == other.isFailableInitializer &&
+        documentation == other.documentation &&
+        annotations == other.annotations &&
+        documentation == other.documentation &&
+        attributes == other.attributes
+    }
+
+    var isDefaultImplementationMethod: Bool {
+        definedInType?.isExtension ?? false
```
cobbal, line 289: This seems like it's either missing cases or is poorly named. Lots of non-default methods are implemented in extensions

Sources/FishyJoesCore/Translators/KotlinTranslator.swift
```diff
@@ -18,7 +18,7 @@ final class KotlinTranslator: Translator {
 
         if let selfType = method.definedIn {
             let resolved = context.resolve(type: selfType)
-            containingNamespace = resolved.sourceType.name
+            containingNamespace = resolved.converterType.name
```
cobbal: This doesn't seem right. "containingNamespace" is supposed to refer to the namespace where the method is defined. converterType is a potentially different type in a potentially different module. Maybe it's being used wrong somewhere, but the fix should be at the usage.

cobbal (commented):
Still working through this review, but commenting with what I have so far. The default implementations worked out much more cleanly than I thought it might.

Sources/FishyJoesCore/Method.swift
dvosscricut: it's for unnamed parameters in a method.

Sources/FishyJoesCore/Method.swift
dvosscricut: good point

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
dvosscricut: agreed. However it is required here, as TestAPI.AProtocol doesn't work.

Sources/FishyJoesCore/TranslatedTypes/TranslatedType.swift
dvosscricut: I think you're right. It should be a defined in extension AND is a protocol type. The bug probably just didn't get exposed by the current TestAPI.

.swiftlint.yml
```diff
@@ -6,6 +6,7 @@ included:
 
 excluded:
     - ./Sources/FishyJoesExecute/Resources
+    - ./integration-tests/TestAPI-bindings/Sources/Generated/FishyJoes.generated.swift
```
cobbal: This file should not exist, and is probably caused by print statements inside of FishyJoesCore. It should definitely not be excluded from lint.

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
```diff
@@ -60,6 +60,28 @@ struct TranslatedProtocol: TranslatedType {
         self.documentation = type.documentation
         self.className = context.kotlinTranslator.javaClassName(kotlinName, in: context)
         self.externalWitnessClassName = context.kotlinTranslator.javaClassName("_ExternalWitness_\(kotlinName)", in: context)
+
+        enforceProtocolThrows()
+    }
+
+    func enforceProtocolThrows() {
+        let nonThrowingMethods = methods.filter { !$0.isThrowing }
+        guard nonThrowingMethods.isEmpty else {
```
cobbal, line 62: nit: a "guard let" is a good way to avoid needing to force-unwrap first

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
```diff
@@ -60,6 +60,28 @@ struct TranslatedProtocol: TranslatedType {
...
+            var nameSpace = ""
+            if let ns = nonThrowingMethods.first!.definedIn?.name {
+                nameSpace = "\(ns)."
+            }
+            fatalError("Error on \(nameSpace)\(nonThrowingMethods.first!.name): All Protocol methods exported through FishyJoes must be throwing, it's the law 👮!")
+        }
+
+        let nonThrowingGetters = computedVariables.filter { !$0.`throws` && $0.writeAccess == AccessLevel.none.rawValue }
```
cobbal: If I understand this logic, this allows for non-throwing getters, but only when we also have a non-throwing setter? I think any non-throws getter or setter will have the same problem.

integration-tests/TestAPI-bindings/Package.swift
```diff
@@ -51,6 +52,10 @@ let package = Package(
                 .unsafeFlags(["-warn-concurrency"])
             ]
         ),
+        .target(
+            name: "CommonInterface",
```
cobbal: This should be named "TestAPI_CommonInterface" for consistency. It also depends on FishyJoesCommonRuntime

cobbal: Actually, it might be easier to make it not conform to `FishyJoesCommonRuntime.Converter` explicitly, but to just define the typealias. The protocol conformance doesn't really buy anything but documentation, and I'd rather not make `FishyJoesCommonRuntime` a public library.

Sources/FishyJoesCore/Method.swift
dvosscricut: good point. Should be named isInExtension.

I don't care if a method is in an extension or not generally, just when it's also a Protocol method, because a protocol method in an extension must be a default implementation. Since Methods are used in Protocol, Class, Struct, and Enum, I need to AND the isInExtension with isProtocol type to get if it's default implementation or not.

Sources/FishyJoesCore/Method.swift
dvosscricut: I need to use it in TranslatedProtocol.swift when looping over protocol methods, so putting it on Method made more sense to me.

Sources/FishyJoesCore/Translators/KotlinTranslator.swift
```diff
@@ -35,21 +35,39 @@ final class KotlinTranslator: Translator {
...
         }
-
-        let formals = [
-            (name: "_javaEnv", type: "UnsafeMutablePointer<JNIEnv?>"),
-            (name: "_javaThis", type: "jobject"),
-        ] + method.parameters.map { parameter in
-            let resolved = context.resolve(type: parameter.type, generics: exportAnnotation.genericOverrides)
-            return (name: parameter.name, type: resolved.converterType.name + ".CType")
+        var formals = [(name: "_javaEnv", type: "UnsafeMutablePointer<JNIEnv?>")]
+        if method.isInExtension {
```
cobbal: This should probably only apply to protocols

Sources/FishyJoesCore/Translators/KotlinTranslator.swift
```diff
@@ -35,21 +35,39 @@ final class KotlinTranslator: Translator {
...
+        }
+        formals.append((name: "_javaThis", type: "jobject"))
+        do {
+            var unnamedParamCnt = 1
+            formals.append(
+                contentsOf:
+                    method.parameters.map { parameter in
+                        let resolved = context.resolve(type: parameter.type, generics: exportAnnotation.genericOverrides)
+                        var name = parameter.name
+                        if name.isEmpty {
```
cobbal: Removing the name seems to create more problems than it solves. This parameter had a perfectly good name, there's no reason to refer to it by a number instead.

Sources/FishyJoesCore/TranslatedTypes/TranslatedType.swift
dvosscricut: well, perhaps

Sources/FishyJoesCore/Translators/KotlinTranslator.swift
```diff
@@ -251,24 +294,50 @@ final class KotlinTranslator: Translator {
                     typeSetupFragment.output("// print(\"setting up \(resolved.converterType.name)...\")")
                     typeSetupFragment.output("try \(resolved.converterType.name).javaSetup(env: env)")
                     if case .named = type {
-                        if let nativeMethods = allMethods[type.name] {
-                            typeSetupFragment.outputBlock("try env.RegisterNatives(\(resolved.converterType.name).javaClass,", closeWith: ")") {
-                                for (method, signature, cName) in nativeMethods {
-                                    let isLast = cName == nativeMethods.last?.cName
-                                    typeSetupFragment.outputBlock("JNINativeMethod(", closeWith: isLast ? ")" : "),") {
+                        if let nativeMethods = allMethods[resolved.converterType.name] {
```
cobbal: I think `allMethods` should be keyed on the source type, not the converter type for reasons similar to the ones mentioned about "containingNamespace"

Sources/FishyJoesCore/Translators/KotlinTranslator.swift
```diff
@@ -251,24 +294,50 @@ final class KotlinTranslator: Translator {
...
+                                                typeSetupFragment.output("signature: bag.add(\"\(signature)\"),")
+                                                typeSetupFragment.output("fnPtr: unsafeBitCast(\(cName), to: UnsafeMutableRawPointer.self)")
+                                            }
+                                    }
+                                    typeSetupFragment.output()
+                                }
+                            }
+                            typeSetupFragment.outputBlock("try env.RegisterNatives(", closeWith: ")") {
+                                typeSetupFragment.output("\(resolved.converterType.name).externalWitnessClass ?? \(resolved.converterType.name).javaClass", newLineTerminated: false)
+                                for (method, signature, cName, _) in normalMethods {
```
cobbal, line 334: I would expect defaulted methods to still be in the witness class.

Sources/FishyJoesCore/Translators/KotlinTranslator.swift
```diff
@@ -430,4 +499,81 @@ final class KotlinTranslator: Translator {
             return nil
         }
     }
+
+    func conform(_ kotlinClass: KotlinClass, to protoName: String, context: FishyJoesContext) {
```
cobbal, line 520: Ideally, we'd use a BetterType instead of a String here, since protocols could come from any module

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
dvosscricut: actually, it does work here. I changed it to use `sourceType.name` here.

dvosscricut: and in one other place. But the other places I tried broke stuff (in filenames), so I left those alone

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
dvosscricut: I just noticed that `hasADefaultImplementationImpl` is never assigned. So the code always calls `_JavaAProtocol_sans_hasADefaultImplementation(wrapped: self).hasADefaultImplementation(x: x, y: y)`. That seems incorrect. 

Your instructive Animal example did not include calling JNI code.

Would you care to give an example of what the code should look like?

Sources/FishyJoesCore/Unparse/KotlinClass.swift
```diff
@@ -345,8 +413,18 @@ class KotlinEnumClass: KotlinClass {
...
-            fields.filter { !$0.isStatic }.forEach { output(field: $0, to: fragment) }
-            methods.filter { !$0.isStatic }.forEach { output(method: $0, to: fragment) }
+            fields.filter { !$0.isStatic }.forEach {
+                // For enums, we are doing a sealed class in Kotlin, not an interface, therefore there is no override on the fields.
+                var unOverride = $0
+                unOverride.isOverride = false
+                output(field: $0, to: fragment)
+            }
+            methods.filter { !$0.isStatic }.forEach {
+                // For enums, we are doing a sealed class in Kotlin, not an interface, therefore there is no override on the methods.
```
cobbal: Looking at the generated code for TestProtocolEnum.kt, I don't see that it conforms to kotlin's `AProtocol` at all. If that's fixed, can these overrides be turned back on?

integration-tests/TestAPI-bindings/kotlin/src/test/kotlin/com/cricut/testapi/ProtocolTests.kt
cobbal, line 1: An important set of tests missing is implementing a swift-defined protocol in kotlin in the test suite. This should be passed to the swift side as an argument of type "AProtocol".

It should probably override one default method to make sure that both overriding works, and using the default implementation works.

I suspect the kotlin-overridden method will not currently be reachable from swift

Sources/FishyJoesJavaRuntime/Env.swift
```diff
@@ -52,6 +56,35 @@ public struct Env {
...
+            self.javaVM = javaVM
+        }
+
+        @inline(__always)
+        private var fns: JavaVM.Pointee {
+            javaVM.pointee!.pointee
+        }
+
+        /// Gets or creates the JNI environment associated with the current thread
+        public func currentThreadEnv() throws -> Env {
```
cobbal: I think the recent async work put a function like this inside of the Env type that more carefully deals with attaching threads. Namely the pair "acquireJVMThread" and "relinquishJVMThread". They should probably either be used directly or wrapped in a block-based convenience method.

Sources/FishyJoesJavaRuntime/JavaConverter.swift
```diff
@@ -10,6 +10,9 @@ public protocol JavaConverter: Converter {
 
     static var javaClass: jclass? { get }
 
+    // only defined for protocols
+    static var externalWitnessClass: jclass? { get }
```
cobbal, line 14: nit: It would be nice not to put this in the JavaConverter protocol, and just have it be on the protocol types. I can't think of a reason we would need to know this dynamically.

Sources/FishyJoesJavaRuntime/JavaReference.swift
```diff
@@ -5,6 +5,10 @@ import JNI
 public class JavaReference {
     public private(set) var object: jobject?
     private let vm: UnsafeMutablePointer<JavaVM?>?
+    @inline(__always)
+    private var fns: JavaVM.Pointee {
+        vm!.pointee!.pointee
+    }
```
cobbal, line 11: This seems to duplicate the functionality in Env.swift

integration-tests/TestAPI-bindings/Sources/Generated/FishyJoes.generated.swift
cobbal, line 1: I think I already mentioned it, but this file is garbage, and indicates print output or code fragments leaking without a proper source fragment set up.

integration-tests/TestAPI-bindings/Sources/Generated/JavaInterface/TestOptionalsProtocol+java.swift
cobbal, line 1: This file (and some other protocol-related ones) are named inconsistently with the rest of the files in the folder. Most start with "TestAPI."

integration-tests/TestAPI-bindings/doIt.sh
```diff
@@ -0,0 +1,6 @@
+rm -f .build/arm64-apple-macosx/debug/libTestAPI.dylib
+rm -f .build/arm64-apple-macosx/debug/libTestAPI-java.dylib
+/Applications/Xcode_15.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/swift-build --product TestAPI-java --configuration debug -j 1 -Xlinker -rpath -Xlinker @loader_path
+cp .build/arm64-apple-macosx/debug/libTestAPI.dylib kotlin/src/generated/resources/mac/libTestAPI.dylib
+cp .build/arm64-apple-macosx/debug/libTestAPI-java.dylib kotlin/src/generated/resources/mac/libTestAPI-java.dylib
```
cobbal: This script should not be committed

integration-tests/TestAPI-bindings/kotlin/build.gradle.kts
```diff
@@ -101,6 +101,11 @@ tasks {
         sourceCompatibility = JavaVersion.VERSION_11.toString()
         targetCompatibility = JavaVersion.VERSION_11.toString()
     }
+    tasks {
+        test {
+            useJUnitPlatform()
```
cobbal: I think this is a duplicate of line 65?

integration-tests/TestAPI-bindings/kotlin/build.gradle.kts
```diff
@@ -111,4 +116,5 @@ dependencies {
     implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core:1.7.3")
     testImplementation("org.jetbrains.kotlinx:kotlinx-coroutines-core:1.7.3")
     testImplementation("org.jetbrains.kotlinx:kotlinx-coroutines-test:1.7.3")
+    testImplementation(kotlin("test"))
```
cobbal: Adding an unversioned dependency is probably not a good idea. What new feature does this give?

integration-tests/TestAPI-bindings/kotlin/src/test/kotlin/com/cricut/testapi/MethodTests.kt
```diff
@@ -26,6 +36,17 @@ internal class MethodTests {
...
+//        println("garply: $garply")
+
+//        val wibble = instance.wibble
+//        println("wibble: $wibble")
+
+        val doublePlusGood = instance.doublePlusGood(34, 4.5)
+        println("doublePlusGood: $doublePlusGood")
+
+        val a = instance.instanceGet
+        print("a: $a")
```
cobbal: These seem more like debugging code left in than tests. Are they still worth having?

scripts/compile-kotlin-native-runtime.sh
```diff
@@ -7,7 +7,7 @@ if [[ ! -d kotlin-runtime ]]; then
     exit 1
 fi
 
-CONFIGURATION="${CONFIGURATION:-release}"
+CONFIGURATION="${CONFIGURATION:-debug}"
```
cobbal: This would turn on debug builds for all our releases. The intended way to get a debug build is to run `CONFIGURATION=debug ./scripts/compile-kotlin-native-runtime.sh`

cobbal (changes requested):
It's looking pretty good. I think there are still some important parts missing though

.swiftlint.yml
dvosscricut: can you give me an example of where these print statements are? I searched for `print(` and only found a lot of comment out code in TypeSetup.swift

until I find out how to disable the generation of this file, SwiftLint will fail on it, unless it's excluded.

Sources/FishyJoesCore/Method.swift
dvosscricut: renaming done (and scoping to only use within SourceryProtocol type

Sources/FishyJoesCore/TranslatedTypes/TranslatedType.swift
dvosscricut: I renamed it and made sure to scope it with the AND is SourceryProtocol when using it to detect default implementations of protocol methods

Sources/FishyJoesJavaRuntime/Env.swift
dvosscricut: this was ported from your protocol branch.

integration-tests/TestAPI-bindings/Package.swift
dvosscricut: this was ported directly from your protocol branch

integration-tests/TestAPI-bindings/Sources/Generated/FishyJoes.generated.swift
dvosscricut: example of such leaky code? I didn't find any print( statements in the code base.

integration-tests/TestAPI-bindings/Sources/Generated/JavaInterface/TestOptionalsProtocol+java.swift
dvosscricut: that they are. I was just trying to get it to work and the TestAPI prefix was messing it up in some of the generated files. But, to be consistent, should be fixed.

Sources/FishyJoesCore/Method.swift
dvosscricut: async and throws supported now added

Sources/FishyJoesCore/Translators/KotlinTranslator.swift
dvosscricut: well, perhaps. It made it work though, and was consistent with your protocols PR.

Sources/FishyJoesExecute/CodeGen.swift
```diff
@@ -164,21 +164,22 @@ extension CodeGen {
...
             // Create / clean directories used by Sourcery to generate Swift and foreign language code files for the translated foreign languages
             let sourceLocations = [
+                "Sources/Generated/CommonInterface",
                 "Sources/Generated/IotaInterface",
                 "Sources/Generated/NodeInterface",
                 "Sources/Generated/JavaInterface",
                 "kotlin/src/generated/kotlin/com/cricut/\(config.module.lowercased())",
                 "c-sharp/Cricut.\(config.module.lowercased())/generated",
                 "dart/lib/src/generated",
+                "integration-tests/TestAPI-bindings/Sources/Generated/FishyJoes.generated.swift"
```
cobbal: Since this is already inside the TestAPI-bindings folder:
```suggestion
                "Sources/Generated/FishyJoes.generated.swift",
```

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
dvosscricut: indeed, updated to use jvalue(pointer:)

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
dvosscricut: Hre is the example:

let java__AProtocolConverter__default_hasADefaultImplementation: @convention(c) (
    UnsafeMutablePointer<JNIEnv?>,
    jobject,
    jobject,
    Swift.Int.CType,
    Swift.Double.CType
) -> Swift.String.CType = { _javaEnv, _javaCompanionThis, _javaThis, x, y in
    FishyJoesJavaRuntime.callbackBody(_javaEnv) { _javaEnv in
        return try Swift.String.toJava(
            try _JavaAProtocol_sans_hasADefaultImplementation(wrapped:
                try _AProtocolConverter.fromJava(_javaThis, env: _javaEnv)
            ).hasADefaultImplementation(
                x: try Swift.Int.fromJava(x, env: _javaEnv),
                y: try Swift.Double.fromJava(y, env: _javaEnv)
            ),
            env: _javaEnv
        )
    }
}

It's now implemented and pushed to remote

.swiftlint.yml
dvosscricut: ha ha ha, I must have made it when I had print statements while de3bugging and committed the file to source control, even though it shouldn't exist for normal code. It's deleted as part of CodeGen now.

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
dvosscricut: It's if it does NOT throw but it is a getter i.e. the writeAccess is .none.

It's detecting any getters that do not throw, and if they exist it will do the police it's the law error.

dvosscricut: I'll just do !$0.isMutable because it's the same thing and also more concise.

Sources/FishyJoesCore/Translators/KotlinTranslator.swift
dvosscricut: indeed!

I have amended to

if isProtocol,
           method.isInExtension {
            formals.append((name: "_javaCompanionThis", type: "jobject"))
        }

integration-tests/TestAPI-bindings/doIt.sh
dvosscricut: fixed!

integration-tests/TestAPI-bindings/kotlin/build.gradle.kts
dvosscricut: I was going to use some deep compare assertion equals function, but ended up not needing to use it.

I'll remove it.

integration-tests/TestAPI-bindings/kotlin/src/test/kotlin/com/cricut/testapi/MethodTests.kt
dvosscricut: well, should probably have an assertion to check value, if it is to be left in

dvosscricut: fixed

scripts/compile-kotlin-native-runtime.sh
dvosscricut: ah, yes, that was when I wanted to make sure I got a debug build so I could step through swift code with lldb. I'll put it back.

integration-tests/TestAPI-bindings/kotlin/build.gradle.kts
dvosscricut: indeed, removed

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
dvosscricut: Since we already know it's not empty, then first must exist. I'm okay with using force unwrap in this case, because it has to be true, and avoids code that will never be executed in the proposed 'guard let first = nonThrowingMethods.first else {`'s block

integration-tests/TestAPI-bindings/Package.swift
cobbal: And at the time it was consistent. Things have been renamed since then. Also, I'm pretty sure now that this is what's causing the parallel build failure.

dvosscricut: OK, it's now named "TestAPI_CommonInterface"

Sources/FishyJoesCore/Method.swift
dvosscricut: if the name property is empty then fishyjoes generates incrementingly named placeholder names such as "_1", "_2", and so forth.

But it really is an unnamed property, so I see it as appropriate to have an empty string for a name of something that's unnamed.

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
dvosscricut: very well, I shall add 'java' to the name

dvosscricut: and, I have!

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
dvosscricut: possibly. I should delete this since we're not trying to handle static methods and variables for protocols

dvosscricut: deleted

by the Cybermen

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
dvosscricut: good point. done (ding!).

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
dvosscricut: look at garply and sqpr functions to see why it's needed.

dvosscricut: well, it's not needed now that I discovered label

Sources/FishyJoesCore/Translators/KotlinTranslator.swift
dvosscricut: @cobbal  Here is the same change in your Kotlin protocols PR:
https://github.com/cricut/FishyJoes/blob/f3b26ce3f0ae58fb238235f3c9d656eefeb5e4bf/Sources/FishyJoesCore/Translators/KotlinTranslator.swift#L21

Screenshot 2024-02-23 at 9 14 23 AM

Sources/FishyJoesCore/FishyJoesContext.swift.orig
cobbal, line 1: This shouldn't be committed. Possibly we should add ".orig" to gitignore, as it looks like it's an artifact of "git mergetool"

Sources/FishyJoesCore/Method.swift
```diff
@@ -135,7 +135,11 @@ extension Method {
 
 extension Method {
     func swiftClosureSignature() -> String {
-        "(\(parameters.map { $0.type.name}.joined(separator: ", ")))\(isAsync ? " async" : "")\(isThrowing ? " throws" : "") -> \(returnType.name)"
+        var sig = "(\(parameters.map { $0.type.name}.joined(separator: ", ")))\(isAsync ? " async" : "")\(isThrowing ? " throws" : "")"
+        if returnType.name != "Void" {
```
cobbal: A closure signature needs the arrow, but a function doesn't. I don't think this method is currently used.

Sources/FishyJoesCore/Translators/KotlinTranslator.swift
dvosscricut: They are in the $Companion class because they exist in the companion object of the Kotlin side protocol

dvosscricut: The external witness implements the Protocol, which has the default implementation in the companion object. So, no need to override it on the ExternalWitness. It's simpler this way, because a default implementation method call on the ExternalWitness will call the companion object on the Protocol.

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
```diff
@@ -0,0 +1,370 @@
...
+                do {
+                    var unnamedParamCnt = 1
+                    for param in method.parameters {
+                        // for protocols, only look at the label (external parameter name), don't use the internal parameter name i.e. name)
+                        var paramName = param.label ?? ""
+                        if paramName.isEmpty {
+                            paramName = "_ _\(unnamedParamCnt)"
+                            unnamedParamCnt += 1
+                        }
+                        paramSigs.append("\(paramName): \(param.type.name)")
```
cobbal: The name is a better name than trying to turn the label into a unique name. This also gets rid of all the ugly "+= 1"s
```suggestion
                        paramSigs.append("\(param.label.map { "$0 " } ?? "")\(param.name): \(param.type.name)")
```

Sources/FishyJoesCore/Translators/KotlinTranslator.swift
dvosscricut: ideally yes. This is a straight port from your previous Kotlin protocols PR: https://github.com/cricut/FishyJoes/blob/f3b26ce3f0ae58fb238235f3c9d656eefeb5e4bf/Sources/FishyJoesCore/Translators/KotlinTranslator.swift#L541

Sources/FishyJoesCore/Unparse/KotlinClass.swift
dvosscricut: oh, good point. I suppose it should if it can, after all.

dvosscricut: ok, it's fixed now.

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
```diff
@@ -0,0 +1,370 @@
...
+                    }
+                }
+                for method in methods {
+                    guard method.name != defaultMethod.name else {
+                        continue
+                    }
+                    fragment.output()
+                    let returnSignature = "\(method.isThrowing ? " throws" : "") -> \(method.returnType.name)"
+                    fragment.outputBlock("public func \(method.name)\(returnSignature) {", closeWith: "}") {
+                        fragment.output("\(method.isThrowing ? "try " : "")wrapped.\(method.callName)(\(method.parameters.map { "\($0.name): \($0.name)" }.joined(separator: ", ")))")
```
cobbal: label should be used here as well. Changing a label in one of the default methods to "_" would show the bug.

Sources/FishyJoesJavaRuntime/JavaConverter.swift
dvosscricut: The externalWitness needs to be declared for all types that can be JavaConvertered, because the env.RegisterNatives in TypeSetup.swift needs it for any type that conforms to a protocol

dvosscricut: I tried putting it in for only TranslatedProtocol in TypeSetup.swift, but got mysterious crashes in JNI unload i.e. TypeSetup at runtime (it built fine though).

Sources/FishyJoesCore/Method.swift
dvosscricut: as discussed, I have fixed this to preserve label and name and deal with the processing of it later, as the _ is the label indeed in the case that it's null.

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
dvosscricut: as discussed priorly, the _sans_ call has been moved to the +javadecl.swift files and the _JavaAProtocol now always calls the JNI method. So if the _sans_ call fails to find a default implementation on the Swift side it will call the java side through the _JavaAProtocol in the Converter.

cobbal (changes requested):
Looking pretty good. Tests are still failing.

This seems to be partly because the TestAPI_CommonInterface has an implicit dependency on TestAPI and FishyJoesCommonRuntime. The FishyJoesCommonRuntime dependency should be removed (delete ": Converter", and remove the import). The TestAPI dependency should be reflected in the Package.swift.

Even if the other languages don't yet support protocols, they should still compile (or it can't be merged).

I do want to get a look at the coverage report, because there are a lot of swift language features interacting in fiddly ways.

Sources/FishyJoesCore/Translators/KotlinTranslator.swift
dvosscricut: labels/names have been fixed

integration-tests/TestAPI-bindings/kotlin/src/test/kotlin/com/cricut/testapi/ProtocolTests.kt
dvosscricut: I added `testProtocolKotlinImpl` in `ProtocolTests.kt` to do this now.

integration-tests/TestAPI-bindings/Sources/Generated/FishyJoes.generated.swift
dvosscricut: ah ha! it was past leaky code that has since been undone. the file needn't exist now!

integration-tests/TestAPI-bindings/Sources/Generated/JavaInterface/TestOptionalsProtocol+java.swift
dvosscricut: well, I just changed it to use the sourceType.name instead of sourceType.nonNameSpacedName for the protocol filenames, and it just worked. How nice!

fixed

Sources/FishyJoesJavaRuntime/Env.swift
dvosscricut: The only currentThreadEnv() that is actually used is in JavaReference (used for javaWitness). 
So I just deleted these on Env.

Sources/FishyJoesCore/FishyJoesContext.swift.orig
dvosscricut: good point. fixed

Sources/FishyJoesCore/Method.swift
dvosscricut: it's not used! deleted!

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
dvosscricut: @cobbal  
except that the name on the protocol may be an empty string in the case of garply on TestMethodsProtocol:
/// 
    func garply(_: String) throws -> String

Using the slightly corrected suggested code above `paramSigs.append("\(param.label.map { "\($0) " } ?? "")\(param.name): \(param.type.name)")`

results in this erroroneus code in TestAPI.TestMethodsProtocol+java.swift

    public func garply(: String) throws -> String {

when it needs to be

 public func garply(_ someAutoGeneratedIncrementingName: String) throws -> String {

dvosscricut: also, names in protocols shouldn't be used, even if they exist. Only labels should. The name (internal parameter name) should be left to the implementer of the protocol. although I believe I do use the internal name if declared on a protocol anyway with TestOptionalsProtocol spqr(_ pipe: AssociatedDataEnum). oh well. I guess I should fix that.

// for protocols, only look at the label (external parameter name), don't use the internal parameter name i.e. name)

In summary, protocols only should declare external interfaces, not any internal interface deets, even though they can (though I don't think it's enforced). Agree?

In any case, in the garply case rather, the name is empty and the label is "_" so we have to come up with auto generated names for this case _1, _2, _3, and so forth.

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
dvosscricut: ah, good catch. I'll make one of the labels of the default implementation test methods be "_" to catch this.

dvosscricut: okay, I made hasADefaultImplementation2 have both unnamed label/name combos, _ name: Type and _: Type to verify this. It's working now for this method signature: `func hasADefaultImplementation2(_ a: String, b: Bool, _: Double) throws -> Double`

Sources/FishyJoesJavaRuntime/JavaReference.swift
dvosscricut: This one is for the javaWitness on the protocol, which needs the vm that the Env doesn't have. And also needs the currentThreadEnv

Sources/FishyJoesCore/Translators/KotlinTranslator.swift
dvosscricut: when I make those 2 recommended converterType to sourceType changes, I get these droppedMethods error:

Users/dvoss/Downloads/code/FishyJoes/Sources/FishyJoesCore/Translators/KotlinTranslator.swift:376: methods for ["TestAPI.TestMethodsProtocol", "TestAPI.AProtocol", "TestAPI.TestOptionalsProtocol", "_TestPropertiesProtocolConverter"] were never set up! Probably a bug.

dvosscricut: also, could we just combine the into the same file. So now we'd just have _AProtocolConverter.swift with all the conversion JNI hookup and to/fromJava stuff all in one file?
 
it seems much less confusing to me that way.

Though maybe the purpose of the separate files it because they are generated in separate functions and we don't want to have the pain of passing the same SourceFragment or matching up between arrays of SourceFragments between the functions. Is that why they're split up?

If that is indeed the case, may I change the name of the +java.swift files to have the converterType.name rather than the sourceType.name? It seems a more logical grouping of for example _AProtocolConverter+javadecl.swift, _AProtocolConverter+java.swift, and _AProtocolConverter.swift rather than having _AProtocolConverter+javadecl.swift, TestAPI.AProtocol+java.swift, and _AProtocolConverter.swift

(as Test.AProtocol+java.swift is actually implementing the fromJava/toJava functions i.e. the conversion functions.

well, I've implemented that now on the PR, swift files containing converting to/fromJava related code are now named under the converterTypeName.

I'll see what feedback you have for me on this decision of mine.

Sources/FishyJoesCore/Translators/DartTranslator.swift
```diff
@@ -107,6 +113,9 @@ final class DartTranslator: Translator {
 
         for nativeMethod in nativeMethods.sorted(by: { $0.name < $1.name }) {
             externDeclarations.append { fragment in
+                if nativeMethod.name.contains("TestProtocolEnum") {
+                    let a = 1
+                }
```
cobbal: A very strange stub

Sources/FishyJoesCore/Translators/KotlinTranslator.swift
```diff
@@ -15,13 +15,15 @@ final class KotlinTranslator: Translator {
...
 
         if let selfType = method.definedIn {
             let resolved = context.resolve(type: selfType)
-            containingNamespace = resolved.converterType.name
+            containingNamespace = resolved.sourceType.name
+            converterNamespace = resolved.converterType.name
 
             if method.isStatic {
-                selfExpression = containingNamespace
+                selfExpression = converterNamespace
```
cobbal: It probably doesn't matter, but this is incorrect. "Self" is containingNamespace, not the converter

Sources/FishyJoesCore/Translators/KotlinTranslator.swift
```diff
@@ -58,10 +61,10 @@ final class KotlinTranslator: Translator {
         }
 
         let fragment = context.swiftFragment(
-            "JavaInterface/\(containingNamespace)+javadecl.swift",
+            "JavaInterface/\(containingNamespace)+java-methods.swift",
             additionalImports: ["Foundation", "FishyJoesJavaRuntime", "\(context.module.name)_CommonInterface"]
         )
-        fragment.output("// Generated by FishyJoes for `\(containingNamespace).\(method.name)`")
+        fragment.output("// Generated by FishyJoes for `\(converterNamespace).\(method.name)`")
```
cobbal: Incorrect documentation. The method is "AProtocol.foo", not "AProtocolConverter.foo"

Sources/FishyJoesCore/Translators/KotlinTranslator.swift
```diff
@@ -298,7 +301,7 @@ final class KotlinTranslator: Translator {
                     if variable.isStatic {
                         fragment.output("\(selfExpression).\(variable.name) = try \(converterName).fromJava(newValue, env: _javaEnv)")
                     } else {
-                        fragment.outputBlock("try \(containingNamespace).mutateJava(_javaThis, env: _javaEnv) { value in", closeWith: "}") {
+                        fragment.outputBlock("try \(destinationRootFilename).mutateJava(_javaThis, env: _javaEnv) { value in", closeWith: "}") {
```
cobbal: Why are we calling methods on a file name?

Sources/FishyJoesCore/TranslatedTypes/TranslatedReference.swift
```diff
@@ -237,7 +237,7 @@ struct TranslatedReference: TranslatedType {
         }
 
         if equatable {
-            context.kotlinTranslator.allMethods[sourceType.name, default: []].append(
+            context.kotlinTranslator.allMethods[converterType.name, default: []].append(
```
cobbal: As mentioned in a previous review:

> I think allMethods should be keyed on the source type, not the converter type for reasons similar to the ones mentioned about "containingNamespace"

Methods are not on the converter. The converter is only for fromJava and toJava. The methods are part of the source type.

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
cobbal: `name` is a much better name than "\_2" will ever be. "garply(\_ str: String)" is infinitely better than "garply(\_ someAutoGeneratedIncrementingName: String)"

Sources/FishyJoesJavaRuntime/JavaReference.swift
```diff
@@ -27,4 +31,21 @@ public class JavaReference {
             self.object = nil
         }
     }
+
+    /// Gets or creates the JNI environment associated with the current thread
+    public func currentThreadEnv() throws -> Env {
```
cobbal: This function should not be added. The async work added in a similar function that keeps better track of the attached threads. (`acquireJVMThread`/`relinquishJVMThread`)

cobbal (changes requested):

Sources/FishyJoesCore/Translators/DartTranslator.swift
dvosscricut: :)

dvosscricut: fixed

Sources/FishyJoesCore/Translators/KotlinTranslator.swift
dvosscricut: fixed

Sources/FishyJoesCore/Translators/KotlinTranslator.swift
dvosscricut: fixed

Sources/FishyJoesCore/Translators/KotlinTranslator.swift
dvosscricut: good question. It's sourceTypeName now

Sources/FishyJoesCore/TranslatedTypes/TranslatedReference.swift
dvosscricut: It's sourceTypeName now. except in the place it must be converterTypeName

but the allMethods is always keyed off of sourceTypeName now.

Sources/FishyJoesJavaRuntime/JavaReference.swift
dvosscricut: how would the acquire/relinquishJVMThread functions be used in this code example?

static var _fooGetMethodID: jmethodID?
    public var foo: String {
        get throws {
            let env = try _javaWitness.currentThreadEnv()
            return try Swift.String.fromJava(
                env.CallObjectMethod(_javaWitness.object, Self._fooGetMethodID),
                env: env
            )
        }
    }

dvosscricut: would this work?

```
static var _fooGetMethodID: jmethodID?
    public var foo: String {
        get throws {
            let env = try Env.acquireJVMThread(on: _javaWitness.vm)
            defer {
                try? Env.relinquishJVMThread(on: _javaWitness.vm)
            }
            return try Swift.String.fromJava(
                env.CallObjectMethod(_javaWitness.object, Self._fooGetMethodID),
                env: env
            )
        }
    }
```

dvosscricut: or perhaps this?

```
static var _fooGetMethodID: jmethodID?
    public var foo: String {
        get throws {
            let env = try Env.acquireJVMThread(on: _javaWitness.vm)
            let fromJava = try Swift.String.fromJava(
                env.CallObjectMethod(_javaWitness.object, Self._fooGetMethodID),
                env: env
            )
            try Env.relinquishJVMThread(on: _javaWitness.vm)
            return fromJava
        }
    }
```

dvosscricut: acquire and relinquish have been implemented

cobbal (approved):

Sources/FishyJoesCore/TranslatedTypes/TranslatedType.swift
dvosscricut: @cobbal  I made an Azure DevOps ticket for it here: https://cricut.visualstudio.com/Cricut/_workitems/edit/478509

Sources/FishyJoesCore/Translators/KotlinTranslator.swift
dvosscricut: https://cricut.visualstudio.com/Cricut/_workitems/edit/478509 ticket to do this

PR #180, opened 2024-03-12
Dvoss/dart tart

dvosscricut:
Implement Simple Protocol Support for Dart


Sources/FishyJoesCore/BetterType.swift
```diff
@@ -214,6 +214,15 @@ extension BetterType {
         }
     }
 
+    var module: String? {
```
cobbal, line 217: This seems incomplete. Generics have modules, I think void has a module (`Swift`).

Sources/FishyJoesExecutionHelper/main.swift
```diff
@@ -25,4 +25,10 @@ guard let context = (DeprecatedNSKeyedUnarchiver.self as UndeprecatedNSKeyedUnar
     fatalError("Something went wrong with executing fishyjoes from sourcery")
 }
 
-print(FishyJoesContext(context: context).translateAll())
+var noStdErrFifo = false
+if CommandLine.argc >= 4,
+      CommandLine.arguments[3] == "--xcode-debug" {
```
cobbal: This is too specific to your configuration. Better would be to just not pass in the `argument["stderrFifo"]` when calling from xcode

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
```diff
@@ -28,7 +29,12 @@ struct TranslatedProtocol: TranslatedType {
         let typeName = exportAnnotation.name
 
         self.sourceType = BetterType(named: type, context: context)
-        self.converterType = .named(.init(name: "_\(sourceType.nonNamespacedName)Converter", module: nil))
+        var module = sourceType.module
+        if module != nil,
+           !module!.isEmpty {
```
cobbal: There are a quite a few issues with this. Too many force unwraps, weird questions about empty module names. Most importantly, this condition isn't ever failed. module will always (I think) end up being "\\(context.module)_CommonInterface"

Sources/FishyJoesCore/TranslatedTypes/TranslatedStruct.swift
```diff
@@ -394,42 +394,65 @@ struct TranslatedStruct: TranslatedType {
...
+        for method in methods {
+            let returnType = context.resolve(type: method.returnType)
+            let commonName = "_\(converterType.genericBaseName.mangledName)_\(method.callName)"
+            let params = method.parameters.map {
+                let resolved = context.resolve(type: $0.type)
+                return "\(resolved.dartType.ffiTag) \($0.name)"
+            }
+            let paramsStr = params.isEmpty ? "" : "\(params.joined(separator: ", ")),"
+            lines.append("typedef \(commonName) = \(returnType.dartType.ffiCreatedTag) Function(\(dartType.ffiUnownedTag) obj, \(paramsStr) OutCreatedRef exn);")
+        }
```
cobbal: I don't think this belongs in TranslatedStruct. There should be nothing specific to structs about how methods are set up

Sources/FishyJoesCore/TranslatedTypes/TranslatedStruct.swift
```diff
@@ -394,42 +394,65 @@ struct TranslatedStruct: TranslatedType {
...
+                    returnVals.append(
+                        .value(
+                            name: "get_\(storedVar.name)",
+                            type: .named(package: nil, name: "ffi.Pointer<ffi.NativeFunction<\(getType)>>")
+                        ) { fragment in
+                            let defaultValue = resolved.dartType.defaultReturnValue.map { ", \($0)" } ?? ""
+                            fragment.output("ffi.Pointer.fromFunction(\(dartType.name()).ffi_get_\(storedVar.name)\(defaultValue)),")
+                        }
+                    )
+                    if storedVar.isPubliclyWritable {
```
cobbal: Since a field of a struct can be mutated by a mutating method, even if the setter isn't public, I think "isPubliclyWritable" is the wrong condition.

Sources/FishyJoesCore/TranslatedTypes/TranslatedType.swift
```diff
@@ -74,7 +74,7 @@ extension TranslatedType {
     }
 
     var mangledName: String {
-        converterType.name.mangled
+        sourceType.name.mangled
```
cobbal, line 77: This will affect lots of things. Why was the change needed?

Sources/FishyJoesCore/Translators/IotaTranslator.swift
```diff
@@ -119,11 +122,25 @@ final class IotaTranslator: Translator {
 
                     mutateBlock {
                         let body = {
-                            fragment.outputBlock("\(selfExpression)\(callName)(", newLineTerminated: false) {
-                                fragment.outputMap(method.parameters, separator: ",") { formal in
-                                    let resolved = context.resolve(type: formal.type, generics: exportAnnotation.genericOverrides)
-                                    return (formal.label.map { "\($0): " } ?? "") +
+                            if method.isInExtension,
+                               let definedInName = method.definedIn?.name {
+                                fragment.outputBlock("try _Iota\(definedInName)_sans_\(method.callName)(wrapped:", closeWith: ")") {
```
cobbal: Are the "sans" wrapper structs language-specific? It might be possible to move them into the common interface and reduce duplication.

Sources/FishyJoesCore/Translators/IotaTranslator.swift
```diff
@@ -119,11 +122,25 @@ final class IotaTranslator: Translator {
 
                     mutateBlock {
                         let body = {
-                            fragment.outputBlock("\(selfExpression)\(callName)(", newLineTerminated: false) {
-                                fragment.outputMap(method.parameters, separator: ",") { formal in
-                                    let resolved = context.resolve(type: formal.type, generics: exportAnnotation.genericOverrides)
-                                    return (formal.label.map { "\($0): " } ?? "") +
+                            if method.isInExtension,
```
cobbal: Like with kotlin, this should be guarded by being a protocol extension, not just an extension

Sources/FishyJoesCore/Unparse/DartClass.swift
```diff
@@ -28,6 +28,7 @@ class DartClass {
         let returnType: DartType
         let deprecation: Deprecation?
         let body: [String]?
+        let isInExtension: Bool
```
cobbal: Is this really a property of a dart method?

Sources/FishyJoesCore/Unparse/DartClass.swift
```diff
@@ -419,14 +444,102 @@ class DartProductClass: DartClass {
         )
     }
 
+    private func ffiFor(fields: [Variable], fragment: SourceFragment) {
+        var peekTypeName = unqualifiedName
+        if isExternalWitness,
+           let conformance = conformances.first {
```
cobbal: This list was sorted alphabetically. First is probably not the correct one.

Sources/FishyJoesCore/Unparse/DartClass.swift
```diff
@@ -419,14 +444,102 @@ class DartProductClass: DartClass {
...
+            peekTypeName = "\(module.name).\(conformance)"
```
cobbal: "module" is the module of the conforming class, not the module of the protocol

Sources/FishyJoesCore/Unparse/DartClass.swift
```diff
@@ -436,8 +549,12 @@ class DartProductClass: DartClass {
                 fragment.outputBlock("static CreatedRef ffi_new(ffi.Pointer ref, OutCreatedRef exn) => check((exn) =>", closeWith: ");") {
                     fragment.output("createRef(\(unqualifiedName)(ref))")
                 }
-            fragment.blankLine()
+                if !conformances.isEmpty {
+                    toStringImpl(fields: fields, fragment: fragment)
```
cobbal: What do conformances have to do with toString?

Sources/FishyJoesCore/TranslatedTypes/TranslatedReference.swift
```diff
@@ -319,12 +319,14 @@ struct TranslatedReference: TranslatedType {
             )
         }
 
+        let (fields, methods) = KotlinClass.separate(fieldsAndMethods: fieldsAndMethods)
```
cobbal, line 322: This pattern keeps repeating. Separate, then just pass both parts anyway. I don't see what the benefit is.

Sources/FishyJoesCore/Translators/DartTranslator.swift
```diff
@@ -112,8 +107,9 @@ final class DartTranslator: Translator {
         }
 
         for nativeMethod in nativeMethods.sorted(by: { $0.name < $1.name }) {
+            let definingDartClass = nativeMethod.definingDartClass + (nativeMethod.isInExtension ? "_DefaultImplementations" : "")
```
cobbal: isInExtension is a terrible name for this variable. Is it about defaulted implementations or extensions? Only in some cases do they align.

Sources/FishyJoesCore/Unparse/DartClass.swift
```diff
@@ -112,16 +117,22 @@ class DartClass {
...
 
-        for field in fields {
-            let baseArgs = field.isStatic ? [] : [thisArg]
-            result["__iota_get_\(field.mangledName)"] = (args: baseArgs, return: field.type)
-            if field.isPubliclyWritable {
-                result["__iota_set_\(field.mangledName)"] = (args: baseArgs + [(field.name, field.type)], return: .void)
+        // Don't add nativeMethods for DartProtocolClass to dartTranslator
+        // They will be handled by the ExternalWitness for that Protocol
+        // except for default Implementations!
+        if !(self is DartProtocolClass) {
```
cobbal: This is a pretty bad use of inheritance. If the fields shouldn't be output in this manner, don't put them into the superclass's "fields" field.

cobbal: Or better yet, The fields that are output in the normal method should be dealt with here, but the fields that are output in a special manner should just go into a field that DartProtocolClass has/deals with.

Sources/FishyJoesCore/Unparse/DartClass.swift
```diff
@@ -135,14 +146,21 @@ class DartClass {
             for param in method.parameters {
                 params.append((param.name, param.type))
             }
-            result["__iota_\(method.mangledName)"] = (args: params, return: method.returnType)
+
+            // Don't add nativeMethods for DartProtocolClass to dartTranslator
+            // They will be handled by the ExternalWitness for that Protocol
+            // except for default Implementations!
+            if !(self is DartProtocolClass) ||
```
cobbal: Again, a superclass should not be testing if it's a subclass. Just use normal overriding mechanisms.

integration-tests/TestAPI-bindings/doIt.sh
cobbal, line 1: Still shouldn't be committed

cobbal (changes requested):

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
```diff
@@ -107,10 +114,105 @@ struct TranslatedProtocol: TranslatedType {
...
+            setupParams.append(
+                .value(
+                    name: method.callName,
+                    type: .named(package: nil, name: "ffi.Pointer<ffi.NativeFunction<\(commonName)>>")
+                ) { fragment in
+                    let defaultValue = resolvedReturn.dartType.defaultReturnValue.map { ", \($0)" } ?? ""
+                    fragment.output("ffi.Pointer.fromFunction(\(sourceType.name)_DefaultImplementations.ffi_\(method.callName)\(defaultValue)),")
+                }
+            )
+        }
```
cobbal, line 188: These 2 loops could be combined

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
```diff
@@ -255,6 +357,220 @@ struct TranslatedProtocol: TranslatedType {
...
+                    }
+                }
+            }
+            fragment.blankLine()
+
+            fragment.outputBlock("public static func mutateIota(_ this: foreignObject, to value: SwiftType, env: Env) throws {") {
+                for computedVar in computedVariables {
+                    let resolved = context.resolve(type: computedVar.typeName.better)
+                    if computedVar.isMutable {
+                        fragment.outputBlock("try env.check { exn in _\(computedVar.name)Setter[env](", closeWith: ")}") {
```
cobbal: This doesn't seem right. mutation shouldn't cause writing of computed variables

Sources/FishyJoesExecutionHelper/main.swift
dvosscricut: as you wish

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
dvosscricut: sourceType.module is optional

boolean expression in if statement can be false and module can be nil

Sources/FishyJoesCore/TranslatedTypes/TranslatedStruct.swift
dvosscricut: dartSetupDelegates exists in the various translated types and do vary in their format between Enum, Protocol, Struct (compare TranslatedEnum+dart.swift starting on line 2 with TranslatedProtocol.swift line starting on line 118. TranslatedType the parent class has a return empty array implementation of this method. If it doesn't belong in TranslatedType subclasses, where should it belong?

Sources/FishyJoesCore/TranslatedTypes/TranslatedStruct.swift
dvosscricut: not in the foreign interface. We only allow getters from dart to swift access for properties/struct fields. Because Swift only supports 'get throws', but does not support 'set throws'

Sources/FishyJoesCore/TranslatedTypes/TranslatedReference.swift
dvosscricut: The benefit is using common code, whereas before the code was duplicated inside the various subclasses of Dart/KotlinClass's init methods, such as DartProductClass, DartEnumClass, and DartProtocolClass.

Also, it's clearer to reason about separated fields and methods at the module boundaries say while debugging than it is with them combined.

Sources/FishyJoesCore/Translators/DartTranslator.swift
dvosscricut: It's the only information I have from Sourcery to determine if a method is a default implementation in a protocol or not. Sourcery named it "isExtension" in SourceryRuntime/Sources/AST/Type.swift . Would that be a better name?

Do you have another name suggestion?

Also, what other criteria do you know of that I could use to detect if a method is a default implementation? If it's in an extension and it's a protocol we're processing, does it not have to be a default implementation?

Sources/FishyJoesCore/Unparse/DartClass.swift
dvosscricut: This is in a computed variable 'nativeMethods'. If it were in a function body I'd agree with you. But the fields need to still exist to output to the ExternalWitness instead of the ProtocolClass. I can't remove them, I need them both to exist, and to filter here. Though I could just not use nativeMethods and use something else for Protocols?

Sources/FishyJoesCore/Unparse/DartClass.swift
dvosscricut: I think I've read about superclasses knowing about subclasses as a bad practice. I'll have to think about this one.

integration-tests/TestAPI-bindings/doIt.sh
dvosscricut: ha ha, how funny that I did that again!

dvosscricut: fixed

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
dvosscricut: do you mean put it all on one line? I think that makes it less readable, and more in line with the parody of swiftness that is put as much as you can on one line with as many chained maps as possible.

dvosscricut: I put it on one line in case that's what you want. If it's not what you want, please give me a code example to help me understand what you'd like. thx!

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
dvosscricut: it's just named wrong. I'll just call it "fields" instead.

Sources/FishyJoesCore/BetterType.swift
dvosscricut: good point.

'tis remedied now

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
dvosscricut: I can avoid using force unwraps at the cost of brevity though

dvosscricut: which I have now pushed to remote, it's now one line longer, not too bad. Is that what you wanted?

Sources/FishyJoesCore/TranslatedTypes/TranslatedReference.swift
dvosscricut: As a silly example, take a module with 10 separate input arguments. Then, instead of passing all 10 input arguments separately, instead combine them in a mass AOrBOrCOrD etc. type enum, then pass that one enum as one argument into the module, then inside the init method of the module, extract all the properties out again. I think you'll agree that passing 10 separate input arguments into the module is easier to reason about.

Add to that AOrBOrCOrD object extracting code being duplicated inside of multiple different module's init methods, and it adds unnecessary code duplication to the list of problems with this approach

Sources/FishyJoesCore/TranslatedTypes/TranslatedStruct.swift
dvosscricut: For example, the `dartSetupParameters` function implementation in "TranslatedReference.swift" (hereafter abbreviated as [TR]) differs from the one in "TranslatedStruct.swift" (Hereafter abbreviated as [TS] in several notable ways:

1)  [TR] has one single hardcoded ForeignSetupParameter, with a single stored variable (the reference), and nothing else in it, and the constructor ffi function name is "ffi_new".
[TS] has a constructor with potentially many storedVariable inputs that can vary, and the constructor ffi function name is "ffi_constructor"

2) [TR] has no other storedVariables than the reference to the swift object.
[TS] can have many storedVariables as well as have methods.

[TE] (TranslatedEnum) doesn't override dartSetupParameters at all, using the default implementation of returning an empty array defined in TranslatedType.swift.

Many other differences can be discovered upon inspection.

Sources/FishyJoesCore/TranslatedTypes/TranslatedType.swift
dvosscricut: It's to name the ffi stuff by the Protocol's name instead of the ProtocolConverter's name. Because the ProtocolConverter only exists on the Swift side, on the Dart side it's just the Protocol or ExternalWitness. There is no ProtocolNameConverter.dart file on the Dart side.

When I do need the converter name such as in type_setup.dart when hooking up external functions to the ProtocolConverter (such as for Loader.shared.once("setup_TestAPI_CommonInterface._AProtocolConverter", () {/'TestAPI_Structs_ReferenceStruct_setup', I grab it explicitly from the type with converterType. Otherwise the sourceType.name should be what it is normally for the mangled name

Sources/FishyJoesCore/Translators/DartTranslator.swift
dvosscricut: How about I call it "isDefaultImplementation" instead? It is only used in TranslatedProtocol, and is hard coded to false for other TranslatedTypes.

I believe if a method is in a Protocol and also in an extension, then it must be a defaultImplementation.

dvosscricut: I believe I'll still need "isExtension" from Sourcery, but I could make a computed property where isDefaultImplementation is based on the Method's definedInType and could only be true if the definedInType is a Protocol type.

dvosscricut: ok, I made an isDefaultImplementation on property. The definedInType way didn't work so I used the isProtocol method instead

Sources/FishyJoesCore/Translators/IotaTranslator.swift
dvosscricut: good point, fixed now

Sources/FishyJoesCore/Translators/IotaTranslator.swift
dvosscricut: no, they aren't unique, the Swift code for the Java sans and the Iota sans are identical except for the struct name. I could indeed move them to a common place.

dvosscricut: What common place/file would you like? +iota-type/+java-type are out since the rest of the files are different between iota and java. same for +iota-methods/+java-methods.

Perhaps in TestAPI_CommonInterface ? As in TestAPI_CommonInterface._AProtocolConverter?

Sources/FishyJoesCore/Unparse/DartClass.swift
dvosscricut: not anymore! It's now "isDefaultImplementation"

Sources/FishyJoesCore/Unparse/DartClass.swift
dvosscricut: I think the more proper way to do it would be to subclass override then nativeMethods In DartProtocolClass and leave the others as a default. I'll do that then, I think.

dvosscricut: done (ding!)

Sources/FishyJoesCore/Unparse/DartClass.swift
dvosscricut: also now done (ding!)

Sources/FishyJoesCore/Unparse/DartClass.swift
dvosscricut: good catch. That's leftover debug code.

It's okay to toString for reference and public constructors. It's okay to display foreign fields. They just need different inputs, fields for reference constructor and storedFields for public constructor.

dvosscricut: fixed (ding!)

Sources/FishyJoesCore/Translators/IotaTranslator.swift
dvosscricut: OK! I did it! I put it in TestAPI_CommonInterface.

Code duplication: reduced!

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
cobbal: sourcetype.module gets set earlier this function (line 31). It's always context.module.

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
cobbal: The methods are being split into normalMethods and defaultMethods, then each is being looped over in nearly identical ways. Instead, just loop over all methods and test inside the loop

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
dvosscricut: not necessarily, if you look inside the BetterType init (line 148), it's possible for module to be nil as returned from this initializer.

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
dvosscricut: oh, I see it now. good point.

dvosscricut: ok, it's fixed now, but I'll wait to push to remote until you're done reviewing (please let me know, thx!)

Sources/FishyJoesExecutionHelper/main.swift
cobbal: Not quite what I had in mind. The command line `fishy-joes` passes an argument to sourcery named "stderrFifo" to tell it to redirect the output. You have added a new command in a different place to override that behavior and not redirect the output. The simpler solution is to just find where you have xcode passing in the sourcery data, and just remove the "stderrFifo" argument from it. No need for this new "noStdErrFifo" variable.

dvosscricut: tricked again! OK I'll change it.

dvosscricut: oh, wait, I do need it there to let the fishy joes emoji executable know if it's being called from the command line or from Xcode.

Sourcery is not rerun when running from Xcode, as I'm running just the emoji fishy joes executable.

Sources/FishyJoesCore/TranslatedTypes/TranslatedStruct.swift
cobbal: Given a swift struct like:

```
/// <!-- FishyJoes.export(Counter) -->
struct Counter {
    public private(set) var count: Int = 0
    /// <!-- FishyJoes.export(tick) -->
    mutating func tick() { count += 1 }
}
```

The real "count" variable lives on the dart side, but the "tick" function lives on the swift side. To call "tick" from dart

1. count is read from the dart "Counter" object
2. a new swift "Counter" is created with with same count
3. The mutating method "tick" is then run on the swift counter
4. The changes are copied back to the original dart "Counter"
 
In order to do step 4, an ffi setter is needed for count. Even though it's not publicly writable.

dvosscricut: good point

dvosscricut: ok, fixed now.

Sources/FishyJoesCore/TranslatedTypes/TranslatedReference.swift
dvosscricut: Additionally, for DartProtocolClass, I can remove the init override entirely because it matches the DartClass init. Code removal and less duplication == win!

Sources/FishyJoesCore/TranslatedTypes/TranslatedStruct.swift
cobbal: There's nothing about adding protocols that should cause only structs to start registering their methods differently. If you delete this block all the tests still pass, so what was it doing?

cobbal (approved):
Once the struct-method-typedef bit is removed, this looks good to merge

Sources/FishyJoesCore/TranslatedTypes/TranslatedStruct.swift
dvosscricut: I was thinking of a potential case that I don't have a test case for yet where somehow a method must need to be called on a struct in that way. But it wouldn't work right now anyways because the typedef isn't even hooked up to anything! ha ha! well, perhaps on a future PR then. 

I made this Excalidraw diagram of the hasADefaultImplementation calls to start with, since I need it for reference to perhaps reason about this potential theoretical test case.

hasADefaultImplementationDartSwiftCalls

PR #183, opened 2024-03-19
Separate out into different files

dvosscricut:
Separate out DartClass.swift (which contains several subclasses including DartEnumClass, DartProductClass, and DartProtocolClass) into separate files per subclass of DartClass:


cobbal (approved):
Thanks for putting this into a new PR

PR #184, opened 2024-03-19
Mutating counter fixes

dvosscricut:
Fix for mutating method in struct 'MutatingCounter' containing a count property with a mutating function 'increment' through 'TestMutatingCounterProtocol' protocol.

Also add Dart side implementation of protocol 'ProtocolDartImpl' 'testProtocolDartImpl' and to test with.

basically needed a fix to handle the case of of mc2 in 'testMutatingCounterProtocol', which is a swift external witness, being able to increment the count correctly and have it reflect on the dart side.


cobbal (commented):
The more I think about mutating protocols, the less I'm certain what to do with them. This PR seems to assume that a protocol is either a TranslatedStruct, or an external witness. There's no great way for a protocol converter to implement "mutateIota" because mutation could mean any number of different things for structs, for enums, etc. Trying and swallowing all errors certainly doesn't seem like the best way forwards.

Semirelatedly, FishyJoes should probably never catch exceptions, except to pass them along to the other side.

dvosscricut:
> The more I think about mutating protocols, the less I'm certain what to do with them. This PR seems to assume that a protocol is either a TranslatedStruct, or an external witness. There's no great way for a protocol converter to implement "mutateIota" because mutation could mean any number of different things for structs, for enums, etc. Trying and swallowing all errors certainly doesn't seem like the best way forwards.
> 
> Semirelatedly, FishyJoes should probably never catch exceptions, except to pass them along to the other side.

FishyJoes must catch the try Box exception in peekIota because it could be that it's not an external witness in which case we must make an IotaReference.

dvosscricut:
> The more I think about mutating protocols, the less I'm certain what to do with them. This PR seems to assume that a protocol is either a TranslatedStruct, or an external witness. There's no great way for a protocol converter to implement "mutateIota" because mutation could mean any number of different things for structs, for enums, etc. Trying and swallowing all errors certainly doesn't seem like the best way forwards.
> 
> Semirelatedly, FishyJoes should probably never catch exceptions, except to pass them along to the other side.

Perhaps only Structs with mutating functions and external witness are the only cases that need mutateIota. Reference and Enum shouldn't need it.

dvosscricut:
> The more I think about mutating protocols, the less I'm certain what to do with them. This PR seems to assume that a protocol is either a TranslatedStruct, or an external witness. There's no great way for a protocol converter to implement "mutateIota" because mutation could mean any number of different things for structs, for enums, etc. Trying and swallowing all errors certainly doesn't seem like the best way forwards.
> 
> Semirelatedly, FishyJoes should probably never catch exceptions, except to pass them along to the other side.

However I could make the ProtocolConverter's mutateIota not have a do catch around the try box but just do the try box so it would pass up the error up the call stack, as it ought to always succeed, as mutateIota ought to only be called on an ExternalWitness for a Protocol.

dvosscricut:
> > The more I think about mutating protocols, the less I'm certain what to do with them. This PR seems to assume that a protocol is either a TranslatedStruct, or an external witness. There's no great way for a protocol converter to implement "mutateIota" because mutation could mean any number of different things for structs, for enums, etc. Trying and swallowing all errors certainly doesn't seem like the best way forwards.
> > Semirelatedly, FishyJoes should probably never catch exceptions, except to pass them along to the other side.
> 
> However I could make the ProtocolConverter's mutateIota not have a do catch around the try box but just do the try box so it would pass up the error up the call stack, as it ought to always succeed, as mutateIota ought to only be called on an ExternalWitness for a Protocol.

I wonder if that's really true though? hmmmm....

dvosscricut:
case we can't support without redefining the mutateIota contract is on git branches

cobb/MutatingCounter or alternatively dvoss/MutatingCounterBreaker.

So, instead we opt to just prohibit mutating functions in protocols entirely.

PR #186, opened 2024-03-20
Enforce no mutating functions in protocols

dvosscricut:
It's the law!

### DIFF
--- a/Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
+++ b/Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
@@ -58,25 +58,26 @@ struct TranslatedProtocol: TranslatedType {
         enforceProtocolThrows()
         enforceNoProtocolSetters()
         enforceNoProtocolStatics()
+        enforceNoProtocolMutatingFunctions()
     }
 
     func enforceProtocolThrows() {
         let nonThrowingMethods = methods.filter { !$0.isThrowing }
         guard nonThrowingMethods.isEmpty else {
             var nameSpace = ""
-            if let ns = nonThrowingMethods.first!.definedIn?.name {
+            if let ns = nonThrowingMethods.first?.definedIn?.name {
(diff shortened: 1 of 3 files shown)


Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
```diff
@@ -58,36 +58,37 @@ struct TranslatedProtocol: TranslatedType {
...
     func enforceProtocolThrows() {
         let nonThrowingMethods = methods.filter { !$0.isThrowing }
         guard nonThrowingMethods.isEmpty else {
             var nameSpace = ""
-            if let ns = nonThrowingMethods.first!.definedIn?.name {
+            if let ns = nonThrowingMethods.first?.definedIn?.name {
                 nameSpace = "\(ns)."
             }
-            fatalError("☠️ Error on \(nameSpace)\(nonThrowingMethods.first!.name): All Protocol methods exported through FishyJoes must be throwing, it's the law 👮!")
+            fatalError("☠️ Error on \(nameSpace)\(nonThrowingMethods.first?.name ?? "?"): All Protocol methods exported through FishyJoes must be throwing, it's the law 👮!")
```
cobbal, line 71: nit: This is what an `if let` is for.

cobbal (approved):

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
dvosscricut: I prefer the terser syntax of ?? "?"

cobbal: It's not terser, and it has a never executed code path

cobbal: To be clear, I'm saying replace the guard with an if let that binds the thing you care about

PR #188, opened 2024-03-21
Trim what we don't need

dvosscricut:
I added ffi for methods even though I only needed it for fields


cobbal (approved):

dvosscricut:
We need the ffiMethods still because a property could become a method through FishyJoes.exportAsMethod and we still need an ffi_set to mutate an iota witness from a method on the swift side for a mutating function in a struct.

PR #191, opened 2024-03-22
Dvoss/microsoft edge cases

dvosscricut:
Fix plus added regression test for Protocols exported to Dart that have a leading underscore in their property names


Sources/FishyJoesCore/Unparse/DartClass.swift
```diff
@@ -235,7 +234,7 @@ class DartClass {
                 func outputParameter(parameter: Method.Parameter) {
                     let labelComment = parameter.labelComment.map { "/* \($0) */ " } ?? ""
                     let defaultValue = parameter.defaultValue.map { " = \($0)" } ?? ""
-                    fragment.output("\(parameter.type.name(in: self)) \(labelComment)\(DartClass.deforbidify(parameter.name))\(defaultValue),")
+                    fragment.output("\(parameter.type.name(in: self)) \(labelComment)\(Self.deforbidify(parameter.name))\(defaultValue),")
```
cobbal: "Self" should be avoided in classes unless you actually want the dynamic type of the current instance. Since there seems to be no overloads of "deforbidify", it should remain "DartClass"

cobbal (changes requested):

mstokercricut (approved):
Great new tests.

Sources/FishyJoesCore/Unparse/DartClass.swift
dvosscricut: ok

cobbal (approved):
Looks good to me if CI passes (nothing windows-specific in here, so that one check can be ignored due to flakiness for now)

PR #192, opened 2024-03-23
Add support for mutating methods on Protocols for Dart and Kotlin

dvosscricut:
Mutation Nation...

Add support for mutating methods on Protocols for Dart and Kotlin.


integration-tests/TestAPI-bindings/Sources/Generated/JavaInterface/TestAPI.TestMutatingCounterProtocol+java-type.swift
```diff
@@ -0,0 +1,113 @@
...
+            jvalue(pointer: Box(value).retainedOpaque())
+        )
+    }
+    public static func mutateJava<R>(_ this: jobject?, env: Env, body: (inout SwiftType) throws -> R) throws -> R {
+        if env.IsInstanceOf(this, AnyBox.javaClass) {
+            let box = try Box<SwiftType>.fromJava(this, env: env)
+            return try body(&box.value)
+        }
+        var javaWitness = _JavaTestMutatingCounterProtocol(_javaWitness: try JavaReference(local: this, env: env)) as SwiftType
+        return try body(&javaWitness)
```
cobbal, line 98: any new value written to "javaWitness" is silently discarded here. This is still the core of the issue. body gets an inout variable, but the "out" is just thrown away.

dvosscricut: Fortunately, we don't need the output of javaWitness, because in the body the mutations occur through calling methods on the javaWitness, which mutates the java side object. In fact, the body could just have an "in" of javaWitness and still work just fine.

cobbal: That argument amounts to "no one is currently using the function that way" which is a dangerous way to develop a library. This function doesn't know what "body" is going to do, other than that it has an inout parameter.

dvosscricut: We could define a contract of what body should do with the inout mutatingSelf parameter.

mutatingSelf is an object that conforms to the TestAPI.TestMutatingCounterProtocol. 

What else would the user do with this object besides call the things defined in the protocol?

If they decided to be wacky and assign something completely new that conforms to the TestMutatingCounterProtocol to mutatingSelf, then yes, the mutations wouldn't happen correctly on the Iota side in that case.

So, just tell the user how to use the function. Moreover, WE are that user, so once we define the contract, the code generator can conform to that contract. 

If an outside user tried to do the wacky for some zany reason, we can just point to the documented contract of 'mutateJava' and say, it doesn't work because you didn't follow the contract.

dvosscricut:
What happened to Protocol Oriented Programming? https://www.youtube.com/watch?v=p3zo4ptMBiQ. Micah uses plenty of protocols in CriCanvas/Graphics. Protocols/Interfaces == Bad architecture? That seems the opposite of what I've heard.
Perhaps, though doing FishyJoes at all has that type of risk as well, which risk we are already undertaking.
Existence of workarounds are good
It's true though that having a complicated edge case based mutation interface would not be clean and problems could creep in. I wanted to try, and succeeding and gaining more insight and knowledge into how things work in FishyJoes at least. I have closed the PR.

PR #194, opened 2024-03-26
Speed up DataConverter for both node native and wasm

cobbal:
Uglier than I thought this would be. A runtime extension was needed for the wasm side of things because of assumptions N-API made about where the memory backing ArrayBuffer lives.


node-runtime/fishyjoes-runtime-common/Runtime.extensions.js
```diff
@@ -1,4 +1,4 @@
-function applyExtensions(library) {
+function applyExtensions(library, { wasmNapi }) {
```
kellyo-cricut: This could have a default value if you didn't want to modify all the other call sites. `applyExtensions(library, { wasmNapi } = {}) { ... }`

kellyo-cricut (approved):

PR #198, opened 2024-03-28
Add Async methods and computed variables in Protocols support to Kotlin

dvosscricut:
Add Async methods and computed variables in Protocols support to Kotlin


Sources/FishyJoesCore/TranslatedTypes/TranslatedFunction.swift
```diff
@@ -65,7 +70,8 @@ struct TranslatedFunction: TranslatedType {
             return TranslatedFunction(
                 parameters: parameters,
                 returnType: translatedFutureReturn,
-                isAsync: false
+                isAsync: false,
+                isThrowing: false
```
cobbal: All function types we export should be marked throwing. Since in js/kotlin/c#/dart all functions are throwing

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
```diff
@@ -583,27 +586,29 @@ struct TranslatedProtocol: TranslatedType {
         let foreignProtocolType = "_Java\(sourceType.nonNamespacedName)"
 
         var methodIDs: [(idHandle: String, name: String, signature: String)] = []
-        let defaultMethods = methods.filter { $0.isDefaultImplementation }
-        let normalMethods = methods.filter { !$0.isDefaultImplementation }
+        let defaultMethods = methods.filter { $0.isDefaultImplementation && !$0.isAsync }
+        let normalMethods = methods.filter { !$0.isDefaultImplementation && !$0.isAsync }
+        let asyncMethods = methods.filter { $0.isAsync }
```
cobbal: This feels like a combinatoric explosion. What about things that are both default and async?

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
```diff
@@ -672,7 +678,9 @@ struct TranslatedProtocol: TranslatedType {
 
             fragment.output("public static var externalWitnessConstructor: jmethodID?")
 
-            if !defaultMethods.isEmpty {
+            let doExternalCompanionClass = !defaultMethods.isEmpty ||
```
cobbal: Should probably just always do the companion. The compiler doesn't mind if it's empty, and it makes our logic simpler.

cobbal: In general in code-gen, I like to keep the 0 cases consistent with the many cases, and not filter them out. Sometimes it's not possible depending on how aggressive downstream compilers are.

Sources/FishyJoesCore/Unparse/DartProtocolClass.swift
```diff
@@ -139,6 +139,10 @@ class DartProtocolClass: DartClass {
                 }
                 fragment.blankLine()
 
+                if unqualifiedName.contains("AProtocol"),
+                   method.name.contains("DefaultImplementation2") {
+                    let elegoo = 1
+                }
```
cobbal: Probably wasn't meant to be committed.

integration-tests/TestAPI/Sources/TestAPI/Protocols.swift
```diff
@@ -247,3 +247,189 @@ public struct TestLeadingUnderscoredPropStruct: TestLeadingUnderscoredProp {
...
+    /// <!-- FishyJoes.export(exercise5) -->
+    func exercise5(_ fn: @escaping AsyncFunctions.AFun5) async throws -> String
+    /// <!-- FishyJoes.export(exercise6) -->
+    func exercise6(_ fn: @escaping AsyncFunctions.AFun6) async throws -> String
+    /// <!-- FishyJoes.export(thunkTwiceMaker) -->
+    func thunkTwiceMaker(thunk: @escaping () async throws -> Void) throws -> () async throws -> Void
+}
+
+/// <!-- FishyJoes.export(TestAsyncForeignSideFunctionsStruct, conformances: [TestAsyncFunctions]) -->
+public struct TestAsyncForeignSideFunctionsStruct: TestAsyncFunctions {
```
cobbal, line 299: I'm not sure this is testing anything fundamental to protocols. We're already testing that kotlin functions can be passed and called in other places.

Sources/FishyJoesCore/BetterType.swift
```diff
@@ -205,8 +210,54 @@ extension BetterType {
...
+                if Int($0.label) == nil {
+                    return "\($0.label): \($0.type.name)"
+                } else {
+                    return $0.type.name
+                }
+            }.joined(separator: ", ") + ")"
+        case .void:
+            return "Void"
+        case let .function(args, ret, isAsync, isThrowing):
+            return "(\(args.map(\.escapingFunctionsName).joined(separator: ", ")))\(isAsync ? " async" : "")\(isThrowing ? " throws" : "") -> \(ret.name)"
```
cobbal: I think I see what you're trying to do with these changes, and I think it can be cleaned up significantly:

```
    var name: String {
        switch self {
        case let .named(name):
            return name.globalName
        case .tuple(let elements):
            return "(" + elements.map {
                if Int($0.label) == nil {
                    return "\($0.label): \($0.type.name)"
                } else {
                    return $0.type.name
                }
            }.joined(separator: ", ") + ")"
        case .void:
            return "Void"
        case let .function(args, ret, isAsync, isThrowing):
            return "(\(args.map(\.escapingName).joined(separator: ", ")))\(isAsync ? " async" : "")\(isThrowing ? " throws" : "") -> \(ret.name)"
        case .generic(let base, let args):
            return "\(base.name)<\(args.map(\.name).joined(separator: ", "))>"
        case .selfType:
            return "Self"
        }
    }

    var nonNamespacedName: String {
        switch self {
        case let .named(name):
            return name.name
        default:
            return name
        }
    }

    var escapingName: String {
        switch self {
        case .function:
            return "@escaping \(name)"
        default:
            return name
        }
    }
```

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
```diff
@@ -282,7 +283,8 @@ struct TranslatedProtocol: TranslatedType {
             fragment.blankLine()
 
             for field in fields {
-                fragment.output("var \(field.name): \(field.typeName.better.name)")
+                let fieldType = field.typeName.actualTypeName?.name ?? field.typeName.better.name
+                fragment.output("var \(field.name): \(fieldType)")
```
cobbal: Any changes involving actualTypeName shouldn't be needed.

Sources/FishyJoesNodeRuntime/NodeReference.swift
```diff
@@ -2,8 +2,10 @@ import Foundation
 
 public final class NodeReference: @unchecked Sendable {
     private var ref: NAPI.Ref
+    public private(set) var env: NAPI.Env
```
cobbal: This should not be stored, as it's thread-specific (sometimes)

cobbal (changes requested):
Looks pretty good. The escapingName adds a lot of stuff, but I think most of it can be removed after fixing "name"

Sources/FishyJoesCore/TranslatedTypes/TranslatedFunction.swift
dvosscricut: whoakay

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
dvosscricut: oh, I should add a test case for that.

I don't have an async defaultMethod test case (yet).

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
dvosscricut: ok, always do the companion then

Sources/FishyJoesNodeRuntime/NodeReference.swift
dvosscricut: luckily I don't need it right now, just when I try to do the nodeWitness will I might need it.

Sources/FishyJoesCore/BetterType.swift
dvosscricut: thanks! done.

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
dvosscricut: I think this was a remnant of from before when I didn't have escaping name set up.

dvosscricut: ok, now that the better.name puts the @escaping in already I don't need the actualTypeName anymore

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
dvosscricut: things that are both default and async will still get put in the companion, they'd just be under the asyncMethods loop rather than the defaultMethods loop. Perhaps I should have 4 cases to be more clear: defaultAsync, default, normalAsync, normal. 2 input bits, 4 output combinations

integration-tests/TestAPI/Sources/TestAPI/Protocols.swift
dvosscricut: Other places are not protocols though

Since protocols are new and different to what currently exists in terms of tests, I believe we ought to exercise this in tests specifically for async functions in protocols passed in from Kotlin.

It also helped me learn the fancy dance of closures in Kotlin tests though.

dvosscricut: I found some significant CodeGen bugs through this scenario, therefore I believe it should stay to make sure they stay fixed.

Sources/FishyJoesCore/Unparse/DartProtocolClass.swift
dvosscricut: indeed!

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
dvosscricut: Ok, I put the test case in for the default implementation async methods in protocol. And it found more bugs! So good thing it's tested now!

Sources/FishyJoesCore/Unparse/KotlinClass.swift
```diff
@@ -159,7 +158,7 @@ class KotlinClass: NestedClass {
                 }
                 if let body = method.body {
                     precondition(compatibilityParameters.isEmpty, "internal error: compatibilityOrder can't be used with a non-native body")
-                    fragment.output(" = \(body)\(method.returnType.toKotlinType)")
+                    fragment.output(" = \(body)\(method.returnType.toKotlinType)\(method.isSuspend ? ".await()": "")")
```
cobbal: This should probably follow the same logic for line 173. Async UInt methods will probably fail to typecheck

Sources/FishyJoesCore/Unparse/KotlinInterface.swift
```diff
@@ -15,6 +15,20 @@ class KotlinInterface: KotlinClass {
                     unstaticked.isStatic = false
                     output(method: unstaticked, to: fragment)
                 }
+                for method in methods {
+                    if !method.name.hasPrefix("_"),
+                       method.isSuspend {
+                        fragment.output("@JvmStatic")
+                        fragment.outputBlock("fun \(method.name)(", newLineTerminated: false) {
```
cobbal: This name should probably get a prefix to avoid ambiguity with the non-static method. Differing only by static-ness and number of arguments feels like a recipe for future developer confusion.

cobbal (changes requested):
Couple fairly minor issues. Coverage seems a little low. There's not really a concrete percentage we go for in this repo, but it might be worth looking at the report and seeing if you're seeing what you expect to be covered.

Sources/FishyJoesCore/Unparse/KotlinClass.swift
dvosscricut: done

Sources/FishyJoesCore/Unparse/KotlinInterface.swift
dvosscricut: it has the prefix "_default". I updated the code to check for "_default" instead of just "_"

dvosscricut:
> Couple fairly minor issues. Coverage seems a little low. There's not really a concrete percentage we go for in this repo, but it might be worth looking at the report and seeing if you're seeing what you expect to be covered.

Sonar cloud doesn't seem to run the Kotlin tests to determine code coverage. It's saying exercise0-6 are uncovered, even though I explicitly cover them in the ProtocolTests.kt. I don't think Sonar cloud knows how to run Kotlin tests to determine if Swift code is covered.

Screenshot 2024-04-02 at 1 46 19 PM


It is the case though for other targets such as Dart and CSharp and Node that it's complaining about, but those tests will come later.

Sources/FishyJoesCore/Unparse/KotlinInterface.swift
dvosscricut: decided, for the deferred wrapper companion object functions we should use the prefix "_deferred"

integration-tests/TestAPI-bindings/kotlin/src/test/kotlin/com/cricut/testapi/ProtocolTests.kt
```diff
@@ -270,9 +323,9 @@ internal class ProtocolTests {
         val h = a.exercise1(b)
         assertEquals(h, "-45")
         val i = a.exercise2 { a: suspend (Long) -> Long, b: suspend (Long) ->
-            Long -> { z: Long ->
-                a(3) + b(3) + z
-            }
+        Long -> { z: Long ->
+            a(3) + b(3) + z
+        }
```
cobbal: That's pretty ugly indentation

Sources/FishyJoesCore/Unparse/KotlinInterface.swift
dvosscricut: done (ding!)

cobbal (approved):

integration-tests/TestAPI-bindings/kotlin/src/test/kotlin/com/cricut/testapi/ProtocolTests.kt
dvosscricut: right you are, I will fix it!

PR #203, opened 2024-04-04
fix: update util package to include isArrayBuffer

kellyo-cricut:
fixes browser runtime issue in wasm-napi where function was undefined


cobbal (approved):
Good catch, thanks

kellyo-cricut:
closing, dependency removed here: https://github.com/cricut/FishyJoes/pull/204

PR #204, opened 2024-04-04
chore: remove npm package dependencies

kellyo-cricut:
process and util shims were causing downstream issues with browser based modules
rather than pin to some magic working versions, code was ported from node-util source

### DIFF
--- a/Sources/FishyJoesNodeRuntime/Templates/wasm-napi.js
+++ b/Sources/FishyJoesNodeRuntime/Templates/wasm-napi.js
@@ -7,7 +7,7 @@ import { Buffer } from 'buffer';
 
 // mostly borrowed from https://github.com/browserify/node-util
 // borrowing removed deps on Node polyfills util/process
-function isArrayBuffer(value) { 
+function isArrayBuffer(value) {
     if (typeof ArrayBuffer === 'undefined') {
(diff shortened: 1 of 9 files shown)


Sources/FishyJoesNodeRuntime/Templates/wasm-napi.js
```diff
@@ -3,9 +3,54 @@
...
+// borrowing removed deps on Node polyfills util/process
+function isArrayBuffer(value) {
+    if (typeof ArrayBuffer === 'undefined') {
+        return false;
+    }
+    return value instanceof ArrayBuffer;
+}
+
+function isTypedArray(value) {
+    return value?.byteLength !== undefined;
```
cobbal: can we add a check to make sure ArrayBuffer returns false here? https://nodejs.org/api/util.html#utiltypesistypedarrayvalue

cobbal (changes requested):

Sources/FishyJoesNodeRuntime/Templates/wasm-napi.js
kellyo-cricut: yep, good catch

cobbal (approved):

PR #206, opened 2024-04-11
Protocol support for C# (including Async) with tests!

dvosscricut:
Protocol support for C# (including Async) with tests!

### DIFF
--- a/Sources/FishyJoesCore/TranslatedTypes/TranslatedStruct.swift
+++ b/Sources/FishyJoesCore/TranslatedTypes/TranslatedStruct.swift
@@ -16,6 +16,7 @@ struct TranslatedStruct: TranslatedType {
     let jniType: JNIType
     let isInhabited: Bool
     let definingModule: Module
+    let implements: [Type]
     let conformances: Set<String>
 
     init(context: FishyJoesContext, type: Type) {
@@ -41,6 +42,7 @@ struct TranslatedStruct: TranslatedType {
         self.documentation = type.documentation
         self.isInhabited = type.isInhabited
         self.definingModule = context.module
+        self.implements = Array(type.implements.values)
         self.conformances = exportAnnotation.conformances
     }
 
--- a/Sources/FishyJoesCore/TranslatedTypes/TranslatedStruct.swift
+++ b/Sources/FishyJoesCore/TranslatedTypes/TranslatedStruct.swift
@@ -16,7 +16,6 @@ struct TranslatedStruct: TranslatedType {
     let jniType: JNIType
     let isInhabited: Bool
     let definingModule: Module
-    let implements: [Type]
     let conformances: Set<String>
 
     init(context: FishyJoesContext, type: Type) {
@@ -42,7 +41,6 @@ struct TranslatedStruct: TranslatedType {
         self.documentation = type.documentation
         self.isInhabited = type.isInhabited
         self.definingModule = context.module
-        self.implements = Array(type.implements.values)
         self.conformances = exportAnnotation.conformances
     }
 
--- a/Sources/FishyJoesCore/TranslatedTypes/TranslatedStruct.swift
+++ b/Sources/FishyJoesCore/TranslatedTypes/TranslatedStruct.swift
@@ -549,11 +549,7 @@ struct TranslatedStruct: TranslatedType {
         }
 
         registerDartClass(context: context)
-
-        // TODO: Handle Protocols
-        if conformances.isEmpty {
-            registerCSharpClass(context: context)
-        }
+        registerCSharpClass(context: context)
 
         return fragment
     }
--- a/Sources/FishyJoesCore/TranslatedTypes/TranslatedStruct.swift
+++ b/Sources/FishyJoesCore/TranslatedTypes/TranslatedStruct.swift
@@ -558,7 +558,7 @@ struct TranslatedStruct: TranslatedType {
 
     func registerCSharpClass(context: FishyJoesContext) {
         let fieldsAndMethods =
-            computedVariables.compactMap { context.cSharp(field: $0, of: self, useNativeName: false, conformances: conformances) } +
+            computedVariables.compactMap { context.cSharp(field: $0, of: self, useNativeName: false) } +
             methods.compactMap { context.cSharp(method: $0, of: self) }
         let (productFields, productMethods) = CSharpClass.separate(fieldsAndMethods: fieldsAndMethods)
 
@@ -569,7 +569,7 @@ struct TranslatedStruct: TranslatedType {
                 name: cSharpType.name,
                 constructor: .`public`(
                     fields: storedVariables.compactMap {
-                        switch context.cSharp(field: $0, of: self, useNativeName: true, conformances: conformances) {
+                        switch context.cSharp(field: $0, of: self, useNativeName: true) {
                         case .method: fatalErr("Can't export a stored variable `\(self.sourceType.name).\($0.name)` as a method")
                         case .variable(let field): return field
                         case nil: return nil
--- a/Sources/FishyJoesCore/TranslatedTypes/TranslatedStruct.swift
+++ b/Sources/FishyJoesCore/TranslatedTypes/TranslatedStruct.swift
@@ -558,6 +558,7 @@ struct TranslatedStruct: TranslatedType {
         let fieldsAndMethods =
             computedVariables.compactMap { context.cSharp(field: $0, of: self, useNativeName: false) } +
             methods.compactMap { context.cSharp(method: $0, of: self) }
+        let (productFields, productMethods) = CSharpClass.separate(fieldsAndMethods: fieldsAndMethods)
 
         context.add(
             cSharpClass: CSharpProductClass(
@@ -573,7 +574,9 @@ struct TranslatedStruct: TranslatedType {
                         }
                     }
                 ),
-                fieldsAndMethods: fieldsAndMethods
+                fields: productFields,
+                methods: productMethods,
+                conformances: conformances
             )
         )
     }
--- a/Sources/FishyJoesCore/TranslatedTypes/TranslatedStruct.swift
+++ b/Sources/FishyJoesCore/TranslatedTypes/TranslatedStruct.swift
@@ -558,7 +558,7 @@ struct TranslatedStruct: TranslatedType {
 
     func registerCSharpClass(context: FishyJoesContext) {
         let fieldsAndMethods =
-            computedVariables.compactMap { context.cSharp(field: $0, of: self, useNativeName: false) } +
+            computedVariables.compactMap { context.cSharp(field: $0, of: self, useNativeName: false, conformances: conformances) } +
             methods.compactMap { context.cSharp(method: $0, of: self) }
         let (productFields, productMethods) = CSharpClass.separate(fieldsAndMethods: fieldsAndMethods)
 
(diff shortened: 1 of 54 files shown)


Sources/FishyJoesCore/TranslatedTypes/TranslatedType.swift
```diff
@@ -14,6 +14,7 @@ protocol TranslatedType {
     var definingTSNamespace: String? { get }
     var isInhabited: Bool { get }
     var containedNamedTypes: [TranslatedType] { get }
+    var implements: [Type] { get }
```
cobbal: I'm assuming this is SourceryRuntime.Type, not Swift.Type. FishyJoes avoids using this type because it is very often missing (especially across modules) and other reasons that I forget. Instead, this should either be BetterType or TranslatedProtocol.

Sources/FishyJoesCore/TranslatedTypes/TranslatedStruct.swift
```diff
@@ -562,6 +560,7 @@ struct TranslatedStruct: TranslatedType {
         let fieldsAndMethods =
             computedVariables.compactMap { context.cSharp(field: $0, of: self, useNativeName: false) } +
             methods.compactMap { context.cSharp(method: $0, of: self) }
+        let (productFields, productMethods) = CSharpClass.separate(fieldsAndMethods: fieldsAndMethods)
```
cobbal, line 561: This is not an improvement. It's just more code that previously was centralized.

Sources/FishyJoesCore/Translators/CSharpTranslator.swift
```diff
@@ -136,6 +131,13 @@ final class CSharpTranslator: Translator {
         var asMethod = false
 
         let isOverride = field.exportAnnotation?.isOverride ?? false
+        var implementsProtocol = false
```
cobbal: should "implementsProtocol" be "isOverride"?

Sources/FishyJoesCore/Unparse/CSharpClass.swift
```diff
@@ -303,42 +320,46 @@ class CSharpProductClass: CSharpClass {
     }
 
     let constructor: Constructor
-    let fields: [Variable]
-    let methods: [Method]
+    let isExternalWitness: Bool
```
cobbal: This should not be part of CSharpClass. Thankfully, it doesn't seem to be used at all and I think it can be safely deleted.

Sources/FishyJoesCore/Unparse/CSharpProtocolClass.swift
```diff
@@ -0,0 +1,19 @@
+class CSharpProtocolClass: CSharpClass {
+    override func output(to fragment: SourceFragment) {
+        document(documentation, fragment: fragment)
+
+        fragment.outputBlock("public interface \(unqualifiedName) {") {
+            for field in fields {
+                fragment.output("public \(field.type.name) \(CSharpClass.deforbidify(field.name)) { get; \(field.isPubliclyWritable ? "set;" : "") }")
```
cobbal: In C# we use getter methods for all computed properties because (among other reasons) it can cause infinite loops in the debugger, as it recursively eagerly evaluates all fields. https://github.com/cricut/FishyJoes/blob/main/Sources/FishyJoesCore/Translators/CSharpTranslator.swift#L150 . We should probably export all protocol variables as methods to avoid similar traps. As a bonus, it nicely sidesteps the "spit out an empty setter, because C# compels us" issue

c-sharp-runtime/src/Utilities.cs
```diff
@@ -25,12 +25,16 @@ public class Utilities {
             }
 
             lock(TrackedHandles) {
-                Console.WriteLine($"=== Begin live GC handles {TrackedHandles.Count} ===");
-                foreach (var ptr in TrackedHandles) {
+                LinkedList<IntPtr> TrackedHandlesClone = new LinkedList<IntPtr>(TrackedHandles);
+                System.Diagnostics.Debug.WriteLine($"=== Begin live GC handles {TrackedHandlesClone.Count} ===");
```
cobbal: I don't think duplicate logging should be committed. It should probably be just using Console.WriteLine, and if the behavior of that needs to be changed it can be. I sent you an example in teams.

cobbal (changes requested):
Nice to see the iota stuff largely unchanged

Sources/FishyJoesCore/TranslatedTypes/TranslatedType.swift
dvosscricut: Yes, it's an array of SourceryRuntime.Type, in fact SourceryProtocol. The reason I need it is I need the actual field names that are declared by the SourceryProtocol in order to determine whether or not to ignore the asMethod way of declaring properties or not. These are not included in BetterType, though I believe they are in TranslatedProtocol, so I could use that.

In a C# Interface, I declare properties as the property name with a getter and setter, and the setter is empty if the property is read only.  Then, anything that implements that interface needs to use the same naming.

This is opposed to declaring a separate GetSomeField and SetSomeField methods for properties, as is GetSomeField is currently declared for get only properties. 

I didn't want to mess with any existing generated names, just future ones so I needed this check (done in CSharpTranslator around line 135.

Sources/FishyJoesCore/TranslatedTypes/TranslatedStruct.swift
dvosscricut: The field/method separation code, previously duplicated in CSharpProductClass and CSharpEnumClass, was once again duplicated in the new CSharpProtocolClass.

It's better to put common code in a common place, such as a method. Additionally, the initializer override in CSharpProtocolClass can be entirely done away with, making the code smaller and cleaner. (Also, Sonar cloud dislikes duplicated code).

And fields, methods, and conformances became common fields on CSharpClass.

It's the same modification that was done for previously done for Dart and passed off.

Sources/FishyJoesCore/Unparse/CSharpClass.swift
dvosscricut: isExternalWitness is used in DartProductClass to know whether or not to generate ffi for storedFields.

ha ha! button used in C#. So yes indeed it can and will be deleted. Like a Cyberman deleting obsolete units.

c-sharp-runtime/src/Utilities.cs
dvosscricut: <high pitched lemming voice when you click on start level in lemmings 2 tribes> "ok!"

Sources/FishyJoesCore/Unparse/CSharpProtocolClass.swift
dvosscricut: well, I wondered why that was that way. Ok then, that could simplify some previous code issues.

I suppose if it's a property with a setter then we use a normal property, right?

dvosscricut: ah but protocol properties have no setters because there's no set throws in swift, only get throws. womp womp.

Sources/FishyJoesCore/TranslatedTypes/TranslatedType.swift
dvosscricut: well, now that all properties on protocols are exported as methods for C#, perhaps I don't need this anymore.

Sources/FishyJoesCore/Translators/CSharpTranslator.swift
dvosscricut: not in this case, as exported properties of an object conforming to a protocol do not require the override notation. But, since our recent discussion about exporting all properties on a protocol as methods, perhaps this no longer matters so much.

dvosscricut: actually, I think I still need it, in order to know if I need to generate the GetSomeProperty on an object conforming to a protocol/interface in C# or not.

dvosscricut: nope, turns out I don't need it. ha ha!

Sources/FishyJoesCore/TranslatedTypes/TranslatedType.swift
dvosscricut: indeed, I don't! And, I have remooved them

Sources/FishyJoesCore/Unparse/CSharpProtocolClass.swift
dvosscricut: experiments yesterday showed that no more debugger crash with dotnet 8, however there is a crash with toString.

The GetSomeField() requirement is a pain for structs implementing protocols/interfaces but I have done it according to your wishes. Technically, C# tradition does dictate for computed e.g. get only properties to use a method with a Get prefix instead. I have made the code generator generate such a method that returns the stored property in the case of a struct implementing a protocol field requirement as a read/write property.

integration-tests/TestAPI-bindings/c-sharp/Cricut.TestAPI/generated/Collections.cs
```diff
@@ -291,14 +291,41 @@ System.Collections.Generic.ISet<nint> setOfInt
         /// <!-- FishyJoes.export(Collections.CollectionHolder) -->
         /// </summary>
         public record CollectionHolder {
+            public System.Collections.Generic.IList<bool> GetBoolArray() {
```
cobbal: Was this intended to be added? I didn't think this type conformed to any protocols

cobbal (changes requested):

integration-tests/TestAPI-bindings/c-sharp/Cricut.TestAPI/generated/Collections.cs
dvosscricut: hmmm... GetSomeProperty should be added for objects conforming to protocols, but no need to add if they don't. I'll have to fix that one, good catch.

dvosscricut: fixed!

Sources/FishyJoesCore/Translators/CSharpTranslator.swift
```diff
@@ -148,6 +148,19 @@ final class CSharpTranslator: Translator {
         }
         let resolved = context.resolve(type: field.typeName.better)
 
+        var isInProtocol = false
+    isInProtocolCheck:
```
cobbal: I don't think this label is necessary

Sources/FishyJoesCore/Translators/CSharpTranslator.swift
```diff
@@ -148,6 +148,19 @@ final class CSharpTranslator: Translator {
...
+        if let implements = field.definedInType?.implements,
+           !implements.isEmpty {
```
cobbal: This condition isn't needed

cobbal (approved):
Code is a little complex, but looks good otherwise

Sources/FishyJoesCore/Translators/CSharpTranslator.swift
dvosscricut: good point

Sources/FishyJoesCore/Translators/CSharpTranslator.swift
dvosscricut: good point

PR #213, opened 2024-05-03
Protocol Support for Node

dvosscricut:
Protocol Support for Node


Sources/FishyJoesCore/FishyJoesContext.swift
```diff
@@ -183,17 +183,17 @@ public class FishyJoesContext {
...
             } else {
-                methodsToTranslateForTypeDict[type] = type.rawMethods
+                methodsToTranslateForTypeDict[type] = (methods: type.rawMethods, protocolName: nil)
             }
         }
 
-        for (type, methods) in methodsToTranslateForTypeDict {
-            let methods = methods.compactMap { Method($0, type: type) }
+        for (type, (methods, protocolName)) in methodsToTranslateForTypeDict {
+            let methods = methods.compactMap { Method($0, type: type, protocolName: protocolName) }
```
cobbal: This new parameter comes directly from type, but type is already being passed in. Could save some complexity by not passing it.

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
```diff
@@ -346,44 +348,207 @@ struct TranslatedProtocol: TranslatedType {
             fragment.blankLine()
 
             for field in fields {
-                fragment.output("var \(field.name): \(field.type.name)")
+                fragment.outputBlock("var \(field.name): \(field.type.name) {") {
+                    fragment.outputBlock("get throws {") {
+                        fragment.output("let env = _nodeWitness.env")
```
cobbal: This isn't safe in the presence of threads (env is only valid for 1 thread). Instead the same machinery that function types use should be used.

cobbal: Looks like that's "syncOnMainThread": https://github.com/cricut/FishyJoes/blob/main/Sources/FishyJoesNodeRuntime/NodeFunction.swift#L108

Sources/FishyJoesNodeRuntime/NodeReference.swift
```diff
@@ -1,9 +1,11 @@
 import Foundation
 
 public final class NodeReference: @unchecked Sendable {
-    private var ref: NAPI.Ref
+    public let env: NAPI.Env
+    public var ref: NAPI.Ref
```
cobbal: `env` should not be added for reasons above. `ref` shouldn't need to be public, this class should hide that detail.

Tests/NAPITests/node-tests/runner.js
```diff
@@ -2,7 +2,7 @@ const { WASI } = require("@wasmer/wasi");
 const { WasmFs } = require("@wasmer/wasmfs");
 
 module.exports.init = async () => {
-  const { NAPI } = await import("../../../../../../../Sources/FishyJoesNodeRuntime/Templates/wasm-napi.js");
+  const { NAPI } = await import("../../../Sources/FishyJoesNodeRuntime/Templates/wasm-napi.js");
```
cobbal: I think this change broke the integration tests. Possibly you were running the tests from an unexpected working directory.

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
```diff
@@ -346,44 +348,207 @@ struct TranslatedProtocol: TranslatedType {
...
+                        fragment.output("let napiValue = try _nodeWitness.value(env: env)")
+                        fragment.output("let \(field.name) = try env.getNamedProperty(napiValue, \"\(field.name)\")")
```
cobbal: The second "field.name" isn't right if this has been renamed on export

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
```diff
@@ -346,44 +348,207 @@ struct TranslatedProtocol: TranslatedType {
...
+
+                        if case let .function(params, resVal, _, _) = field.type {
```
cobbal: This is very odd. Types shouldn't need to be special cased here.

integration-tests/TestAPI-bindings/node-test/Protocols.test.ts
```diff
@@ -0,0 +1,520 @@
...
+        );
+    }
+
+    hasADefaultImplementation2(a: string, b: boolean, c: number): number {
+        return Number(a) + (b ? 3.3 : 4.4) + (c * 19);
+    }
+}
+
+test('testProtocolEnum', () => {
+    TestAPI.TestProtocolEnum.foo("qux");
```
kellyo-cricut, line 91: What are these calls on 89/91 checking?

integration-tests/TestAPI-bindings/node-test/Protocols.test.ts
kellyo-cricut, line 1: Nit- almost all of the `let` in here could be `const`

integration-tests/TestAPI-bindings/node-test/Protocols.test.ts
dvosscricut: these are functions that return void, but you can see them printing out stuff on the console. They just have a print statement in the body.

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
```diff
@@ -346,44 +348,207 @@ struct TranslatedProtocol: TranslatedType {
...
             }
 
             fragment.blankLine()
+
             fragment.output("@available(*, deprecated, message: \"Not actually deprecated, but this silences warnings because it may refer to deprecated methods\")")
             fragment.outputBlock("public static func nodeSetup(env: NAPI.Env, module: NAPI.Value) throws {") {
-                // TODO
+                // fragment.output("print(\"setting up \(sourceType.name)\")")
+
+                fragment.outputBlock("let fromCoreClass = try NodeClass(") {
```
cobbal: This doesn't really need to be a full class. A function inside an object would be more idiomatic

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
```diff
@@ -346,44 +348,207 @@ struct TranslatedProtocol: TranslatedType {
...
+                                    fragment.output("let env = env.env")
+                                    fragment.output("let global = try env.getGlobal()")
+                                    fragment.output("let object = try env.getNamedProperty(global, \"Object\")")
+                                    fragment.output("let create = try env.getNamedProperty(object, \"create\")")
+                                    fragment.blankLine()
+                                    fragment.output("let createdCore = try env.callFunction(global, create, [coreArg])")
+                                    fragment.blankLine()
+
+                                    let defaultMethods = methods.filter { $0.isDefaultImplementation }
+                                    if !defaultMethods.isEmpty {
```
cobbal: unneeded if statement. Embrace zero as a useful number.

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
```diff
@@ -346,44 +348,207 @@ struct TranslatedProtocol: TranslatedType {
...
+                            fragment.outputBlock(".method { env, info in", closeWith: "},") {
+                                fragment.outputBlock("FishyJoesNodeRuntime.callbackBody(env, info, name: \"fromCore\", expectedArgumentCount: 1, hasNamedOptions: false) { env in", closeWith: "}") {
+                                    fragment.output("let coreArg = try env.argument(at: 0)")
+                                    fragment.blankLine()
+                                    fragment.output("let env = env.env")
+                                    fragment.output("let global = try env.getGlobal()")
+                                    fragment.output("let object = try env.getNamedProperty(global, \"Object\")")
+                                    fragment.output("let create = try env.getNamedProperty(object, \"create\")")
+                                    fragment.blankLine()
+                                    fragment.output("let createdCore = try env.callFunction(global, create, [coreArg])")
```
cobbal: createdCore is perhaps a bad name for this, since it's the full protocol object, created from the core. Maybe "result" would be a better name?

cobbal: The first argument should probably be `object`, since that's what javascript would pass in when calling `Object.create(...)`

Sources/FishyJoesCore/TranslatedTypes/TranslatedReference.swift
```diff
@@ -36,6 +37,17 @@ struct TranslatedReference: TranslatedType {
         self.cSharpType = .named(package: context.module.cSharpNamespace, name: exportAnnotation.cSharpName)
         self.dartType = .named(package: context.module.dartNamespace, name: context.dartTranslator.fakeNamespace(exportAnnotation.name))
         self.methods = type.methods.compactMap { Method($0, type: type) }
+
+        var nodeDefaultMethods = [Method]()
+        // Node objects needs default methods in order to conform to the interfaces it implements, though as an optional method that if the user does not implement, will use the defined default method constant.
```
cobbal: I don't understand why "TranslatedReference" would need to chage to handle default methods, and the comment here doesn't really explain why references would implement protocols differently from all other types

Sources/FishyJoesCore/TranslatedTypes/TranslatedReference.swift
```diff
@@ -158,9 +171,12 @@ struct TranslatedReference: TranslatedType {
             .init(
                 documentation: documentation,
                 name: nodeName,
+                implements: Array(conformances).sorted(by: <),
                 constructor: .hidden,
                 fields: computedVariables.compactMap { context.ts(field: $0) },
-                methods: methods.compactMap { context.ts(method: $0) }
+                methods:
+                    methods.compactMap { context.ts(method: $0) } +
+                defaultMethodsForNode.compactMap { context.ts(method: $0) }
```
cobbal: This is really ugly indentation. Is this really what Xcode recommends?

Sources/FishyJoesCore/Unparse/TypeScriptAnnotations.swift
```diff
@@ -358,7 +375,21 @@ extension TypeScriptAnnotations {
...
+//                                        defaultMethods.forEach {
+//                                            guard let protocolName = $0.protocolName else {
+//                                                fatalError("protocolName required for default method in node for constructor.")
+//                                            }
+//                                            fragment.output("this.\($0.name) = this.\($0.name) || \(protocolName)Defaults.\($0.name)")
+//                                        }
+//                                    }
+//                                } else {
+//                                    fragment.blankLine()
+//                                }
```
cobbal: Commented code should be removed or explained.

Sources/FishyJoesCore/Unparse/TypeScriptAnnotations.swift
```diff
@@ -370,6 +401,9 @@ extension TypeScriptAnnotations {
                             }
                         }
                     case .interface(let interface):
+                        if !interface.documentation.isEmpty {
```
cobbal: unnecessary if

Sources/FishyJoesCore/Unparse/TypeScriptAnnotations.swift
```diff
@@ -272,17 +280,14 @@ extension TypeScriptAnnotations {
             fragment.output(" */")
         }
 
-        func output(method: Method, inClass: Bool) {
+        func output(method: Method, inClass: Bool, optionalMethodsForDefaults: Bool = false, isCore: Bool = false) {
```
cobbal, line 283: Feels like there are a lot more options expressible than combinations of options used. I'm not sure what the right combination is and whether it belongs in the TSType.Method class or as function arguments.

Sources/FishyJoesCore/Translators/NodeTranslator.swift
```diff
@@ -79,7 +79,7 @@ struct NodeTranslator: Translator {
         }
     }
 
-    func output(method: Method, explicitThis: Bool, context: FishyJoesContext, fragment: SourceFragment, newLineTerminated: Bool = true) {
+    func output(method: Method, explicitThis: Bool, context: FishyJoesContext, fragment: SourceFragment, newLineTerminated: Bool = true, converterName: String? = nil) {
```
cobbal: I'm seeing a lot of the defaulted stuff on methods in this PR, but not much on properties. There may also be TS limitations for those: https://github.com/microsoft/TypeScript/issues/36883

integration-tests/TestAPI-bindings/Sources/Generated/NodeInterface/TestAPI.AProtocol+node.swift
```diff
@@ -10,33 +10,235 @@ import TestAPI_CommonInterface
...
+                    throw JSException(message: "expected TestAPI.AProtocol, got nil")
+                }
+                return try Box<TestAPI.AProtocol>.takeUnretainedOpaque(nonNilPointer).value
+            } else {
+                return _NodeAProtocol(
+                    _nodeWitness: try NodeReference(env: env, value: value)
+                )
+            }
+        } catch {
+            throw error
```
cobbal: Is this necessary?

integration-tests/TestAPI-bindings/node-test/Protocols.test.ts
dvosscricut: @kellyo-cricut nit has been fixed

integration-tests/TestAPI-bindings/c-sharp/Cricut.TestAPI/generated/TestProtocolStruct.cs
```diff
@@ -31,52 +31,52 @@ string Corge
         private static extern CreatedRef __iota_get_TestAPI_TestProtocolStruct_frob(IntPtr envRef, UnownedRef self, out CreatedRef exn);
 
         /// <summary>
-        /// <!-- FishyJoes.export(foo) -->
+        /// <!-- FishyJoes.export(bar) -->
         /// </summary>
-        public void Foo(
+        public bool Bar(
```
cobbal: Did the order here become non-deterministic (bad) or is this a one-time change (fine)?

cobbal (changes requested):
Looking pretty good. Structure looks solid at this point.

integration-tests/TestAPI-bindings/node-test/Protocols.test.ts
dvosscricut: If I knew how to test verify what's going to stdout on the console, then I could verify them. Otherwise, I just manually verify them by looking at the console and checking that it prints out as it ought.

Sources/FishyJoesCore/FishyJoesContext.swift
dvosscricut: good point, fixed now

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
dvosscricut: ok, fixed now

Tests/NAPITests/node-tests/runner.js
dvosscricut: ok, I put it back.

Sources/FishyJoesNodeRuntime/NodeReference.swift
dvosscricut: ok, fixed now

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
dvosscricut: ok

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
dvosscricut: what about the case of a field exported as a method? Would I not need to check for that here?

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
dvosscricut: good point

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
dvosscricut: renamed createdCore to result

dvosscricut: made first arg object instead of global

Sources/FishyJoesCore/TranslatedTypes/TranslatedReference.swift
dvosscricut: indeed this is the result of CTRL+I

Sources/FishyJoesCore/Unparse/TypeScriptAnnotations.swift
dvosscricut: cruft uncruftified sir

Sources/FishyJoesCore/Unparse/TypeScriptAnnotations.swift
dvosscricut: ok

Sources/FishyJoesCore/Translators/NodeTranslator.swift
dvosscricut: There aren't default implementations on Swift protocols for properties though, just for methods.

integration-tests/TestAPI-bindings/Sources/Generated/NodeInterface/TestAPI.AProtocol+node.swift
dvosscricut: no, it's just a leftover from debugging. fixed now

integration-tests/TestAPI-bindings/c-sharp/Cricut.TestAPI/generated/TestProtocolStruct.cs
dvosscricut: it's a one time thing. You can check this by running fishy joes multiple times and observing no file changes.

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
dvosscricut: for example, const42, add3Things, fifthThing etc. on TestAsyncFunctions?

These are fields/properties/variables that contain closures.

It just so happens that if I just use the field it generates an AsyncFunctionXConverter<...>.fromNode that works in tests, so I can get rid of it, but I didn't know that I could!

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
dvosscricut: ok, I did it create an object with a function on it way now

Sources/FishyJoesCore/Translators/NodeTranslator.swift
dvosscricut: I needed the special stuff for handling the default implementations of methods, so that's why the special stuff isn't on properties.

Sources/FishyJoesCore/Unparse/TypeScriptAnnotations.swift
dvosscricut: optionalMethodsForDefaults and isCore are options only used for default implementations of functions for generating the typescript interfaces for them. If you disable them (set them to false, you just get the function definitions you had before

basically it's should we put a question mark after the method name, or not.
And should we output an extra "this: " first argument for the method, or not.

It seems to me it belongs in this functions args, not the Method struct itself, because only Node needs it. The other languages don't.

Sources/FishyJoesCore/TranslatedTypes/TranslatedReference.swift
dvosscricut: TranslatedReference implements node default methods similarly to TranslatedStruct. Perhaps TranslatedEnum should as well? But the rest of the TranslatedTypes (well, besides TranslatedProtocol of course) don't.

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
dvosscricut: so, I got rid of the function case for you. Now all fields are handled the same

Sources/FishyJoesCore/Translators/NodeTranslator.swift
cobbal: > There aren't default implementations on Swift protocols for properties though, just for methods.

Where does this restriction come from? Swift allows defaulted properties just fine.

integration-tests/TestAPI-bindings/node-test/Protocols.test.ts
cobbal: These should be fixed at some point. A test suite that relies on looking at the console is not a very good test suite.

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
cobbal: This should also be in the test suite

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
```diff
@@ -442,56 +423,55 @@ struct TranslatedProtocol: TranslatedType {
...
-                                                fragment.output("try env.setNamedProperty(createdCore, \"\(method.callName)\", \(method.callName)Function)")
-                                            }
-                                            fragment.blankLine()
-                                        }
-                                    }
-
-                                    fragment.blankLine()
-                                    fragment.output("return createdCore")
-                                }
+                fragment.outputBlock("let coreObject = try env.createFunction(") {
```
cobbal: This doesn't need to be a function at all. Just use `env.createObject`

Sources/FishyJoesCore/TranslatedTypes/TranslatedReference.swift
cobbal: How does a struct have default methods either? This just doesn't make any sense.

Sources/FishyJoesCore/Unparse/TypeScriptAnnotations.swift
cobbal: I was referring to `TSType.Method` specifically, not to the other `Method` type(s)

Sources/FishyJoesCore/Translators/NodeTranslator.swift
dvosscricut: Swift allows default values for computed properties, but not stored properties. I suppose we could add support for default computed properties in Swift protocols, but I think that should be on a separate PR/PBI, because none of the other languages we've done for Protocol Support support default computed properties either.
Screenshot 2024-05-08 at 2 35 47 PM

integration-tests/TestAPI-bindings/node-test/Protocols.test.ts
dvosscricut: well, perhaps, but how about on a separate PR/PBI? Because none of the other languages that have Protocol support have verified the console output/functions that return void either.

Sources/FishyJoesCore/Translators/NodeTranslator.swift
cobbal: Sounds like a bug in the implementation for other languages

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
dvosscricut: good catch! I had to fix it for kotlin, dart, and c# also (done on this PR!

Sources/FishyJoesCore/Unparse/TypeScriptAnnotations.swift
dvosscricut: I'm not sure what you mean. There's a type: TSType on the Parameters of a method, but there's no TSType.method, just named, optional, union, exactString, and promise.

However, the function args do work.

If you decide what you'd like better later we could always change it.

Since it's currently undecided, could the function args pass for now?

Sources/FishyJoesCore/Translators/NodeTranslator.swift
dvosscricut: indeed it could be a missing feature for full simple protocol support.

But, may I please do it on a separate PBI ticket and PR (since it needs to be done for all exported languages)?

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
dvosscricut: good point

integration-tests/TestAPI-bindings/node-test/Protocols.test.ts
dvosscricut: I made a PBI ticket for it here: https://cricut.visualstudio.com/Cricut/_workitems/edit/476434

Sources/FishyJoesCore/Translators/NodeTranslator.swift
dvosscricut: I made a ticket for it here: https://cricut.visualstudio.com/Cricut/_workitems/edit/476436

Sources/FishyJoesCore/Unparse/TypeScriptAnnotations.swift
cobbal: I apologize, I was misremembering the organization. I meant `TypeScriptAnnotations.Method`, which is TS-specific

Sources/FishyJoesCore/TranslatedTypes/TranslatedReference.swift
dvosscricut: The TranslatedReference/Struct 'defaultMethodsForNode' is for the Type.d.ts definitions.

It's how we did it for default implementations in type script to work for users conforming to AProtocolCore

```
interface AProtocol {
  hasADefaultImpl(...
}

interface AProtocolCore {
  hasADefaultImpl?(...
}
namespace AProtocol {
  function fromCore(core: AProtocolCore): AProtocol
}

// generated class conforming to interface
export class AProtocolImpl implements AProtocol {
  hasADefaultImplementation(...
}

// user defined class conforming to interface
class TestCoreClass implements TestAPI.AProtocolCore {
  // don't need to define default method here since for AProtocolCore the method is optional
}

// to actually use the class need to run it through .fromCore function:
const testCoreClass = new TestCoreClass("Gitang", false);
 const testClass = TestAPI.AProtocol.fromCore(testCoreClass);
```

now can call `testClass.hasADefaultImpl(...` and have it work.

Now for generated classes I could have them conform to AProtocolCore instead of AProtocol, and then I wouldn't need to put the hadADefaultImpl functions in their definition in Type.d.ts, but then users would have to run them through the AProtocol.fromCore method to actually use them. Is that what we really want?

Or perhaps we'd want some custom constructor function on the generated classes conforming to the protocol, that automatically returns the AProtocol.fromCore conforming to AProtocol instead of just itself conforming to AProtocolCore?

dvosscricut: perhaps I should name them "defaultMethodsForNodeTypeDefinition" instead?

I don't have access to the type.implements  later in the code so that's why I got them here and stored them in a property.

dvosscricut: would you prefer something more like this in the Type.d.ts?

```
class AProtocolImplementation implements AProtocolCore {
        constructor(foo: string, baz: boolean)
        foo: string;
        baz: boolean;

        withDefaultMethods(): AProtocol
        /**
         * <!-- FishyJoes.export(bar) -->
         */
        bar(
            x: number,
            y: number
        ): TestAPI.AProtocol;
    }
```

Then to use the generated class you'd need to do 

```
let a = AProtocolImplementationCore("Foo!", false).withDefaultMethods()
a.hasADefaultImplemention(...
```

would that be better?

integration-tests/TestAPI-bindings/node-test/Protocols.test.ts
dvosscricut: Due to Andrew's decision that testing stdout output is undesirable, and the difficulty in doing so

I will remove the print to stdout part of the methods that do so by wrapping it in a static property definition so that I can still see them if I want but other people won't have to.

The test for a method that returns void will now just be that the method is called and no crashes happen such as would happen if the method did not exist or if there was a name mismatch.

Additionally, I could add tests that have an inout arg and/or callback as an arg to methods.

dvosscricut: I have minimized stdout to just the pid printout (useful for attaching to in Xcode for node/Swift dual debugging), with the rest of the prints to stdouts being behind a constant variable check.

Sources/FishyJoesCore/TranslatedTypes/TranslatedReference.swift
dvosscricut: glad we had that chat.

agreed, methods should just carry default methods with no duplicates in case generated struct/reference implemented a default method on the swift side. I should add a test case to test for this.

The only thing that needs to know if a method is default or not is TranslatedProtocol.

agreed, logic to get the no duplicate list of methods from input methods and protocol default implementation methods should be a common logic function.

Sources/FishyJoesCore/Unparse/TypeScriptAnnotations.swift
dvosscricut: oh ok. apology accepted. so optionalMethodsForDefaults is per interface, not per method, so I'd vote for it to remain a function argument.

similarly, the isCore pertains to the interface, not the method itself per se. So remaining function arguments it is!

Sources/FishyJoesCore/TranslatedTypes/TranslatedReference.swift
dvosscricut: ok I have implemented this as we discussed.

Sources/FishyJoesCore/Method.swift
```diff
@@ -122,6 +122,57 @@ extension Method {
...
+    static func methods(type: Type) -> [Method] {
+        var defaultMethods = [Method]()
+        let protocols = type.implements.values.compactMap { $0 as? SourceryProtocol }
+        for prot in protocols {
+            defaultMethods.append(contentsOf: prot.defaultMethods().compactMap { Method($0, type: prot, protocolName: prot.name) })
+        }
+
+        let normalMethods = type.methods.compactMap { Method($0, type: type) }
+
+        let methods = Method.methodsPreferring(.defaultImplementation, methods: normalMethods + defaultMethods)
```
cobbal: Why does this prefer the default implementation? Shouldn't the overriden implementation be preferred in most contexts?

Sources/FishyJoesCore/TranslatedTypes/TranslatedType.swift
```diff
@@ -268,6 +240,45 @@ extension SourceryProtocol {
 }
 
 extension SourceryMethod {
+    enum MethodTypePreference {
+        case defaultImplementation
+        case normal
+    }
+    static func methodsPreferring(_ preference: MethodTypePreference, methods: [SourceryMethod]) -> [SourceryMethod] {
```
cobbal, line 247: Do we really need both this and the one in Method?

cobbal (changes requested):
Things are looking good. 1 issue, 1 question

Sources/FishyJoesCore/Method.swift
dvosscricut: good catch! you are right, for this method (used for everything but Protocols), the normal method should override the default method.

It turns out to not make in any difference in the code generation, but in principle it makes a difference to make sense rather than nonsense. fixed.

Sources/FishyJoesCore/TranslatedTypes/TranslatedType.swift
dvosscricut: In Theory, no. I'm having a dilly of a time just trying to get it to work using just one of them though.

cobbal (approved):

PR #217, opened 2024-05-23
kotlin single repo

cobbal:


Sources/FishyJoesExecute/DockerContext.swift
```diff
@@ -62,6 +82,40 @@ struct DockerContext {
...
+        else {
+            fatalError("Could not parse git credentials")
+        }
+
+        try swsh.cmd("rm", "-rf", ".secrets/git-credentials").run()
+        try swsh.cmd("mkdir", "-p", ".secrets/").run()
+        try swsh.cmd("chmod", "go-rwx", ".secrets").run()
+        try swsh.cmd("cat")
+            .input("\(proto)://\(username):\(password)@\(host)\n")
+            .output(overwritingFile: ".secrets/git-credentials")
```
mstokercricut, line 118: Scary. Well, the files go away after a run on a remote runner, right...?

Sources/FishyJoesExecute/PackageInit.swift
```diff
@@ -62,17 +75,28 @@ public struct PackageInit: ParsableCommand {
         }
     }
 
-    func shouldIgnore(_ fileName: String) -> Bool {
+    enum InstallBehavior {
```
mstokercricut, line 78: Good addition

Sources/FishyJoesExecute/Platform.swift
```diff
@@ -550,35 +558,35 @@ enum Platform: CustomStringConvertible, Hashable {
 
     func outputDir(_ config: FishyJoesConfig) -> String {
         switch self {
-        case .wasm, .node: return "output\(ps)\(platform)"
+        case .wasm, .node: return "bindings/generated/node/packages/\(platform)"
```
mstokercricut, line 561: Did you mean for this to be "bindings/node/packages/\(platform)/generated"? As is, the "generated" directory inside "bindings" seems to be for generated files for all bindings, not those for node or wasm.

documentation/future-ideas/unified-middle.md
```diff
@@ -0,0 +1,16 @@
+# A unified runtime?
```
mstokercricut, line 1: Huh. That would be powerful...

Sources/FishyJoesExecute/CodeGen.swift
```diff
@@ -160,42 +168,180 @@ extension CodeGen {
...
-                "Sources\(ps)Generated\(ps)CommonInterface",
-                "Sources\(ps)Generated\(ps)WasmMainShim",
-                "Sources\(ps)Generated\(ps)NodeNativeShim",
-                "Sources\(ps)Generated\(ps)NodeInterface",
-                "Sources\(ps)Generated\(ps)JavaInterface",
-                "Sources\(ps)Generated\(ps)IotaInterface",
-                "kotlin\(ps)src\(ps)generated\(ps)kotlin\(ps)com\(ps)cricut\(ps)\(config.module.lowercased())",
-                "c-sharp\(ps)Cricut.\(config.module.lowercased())\(ps)generated",
-                "dart\(ps)lib\(ps)src\(ps)generated",
+            let generatedSwiftTargets = ["DummyMain", "IotaInterface", "NodeInterface", "JavaInterface", "CommonInterface"]
```
mstokercricut: Is this actually a "DummyMain" or does it contain statements? I would expect a dummy main to be empty. If it is analogous to WasmMainShim, etc, then "DummyMain" is misleading...

Sources/FishyJoesExecute/CodeGen.swift
```diff
@@ -160,42 +168,180 @@ extension CodeGen {
...
+            let sourceLocations = generatedSwiftTargets.map {
+                "\(swiftBindingsRoot)/Sources/\($0)"
+            } + [
+                "bindings/kotlin/generated/src/main/kotlin/com/cricut/\(config.module.lowercased())",
+                "bindings/c-sharp/generated/Cricut.\(config.module.lowercased())",
+                "bindings/dart/generated/lib/src",
             ]
-            try cmd("rm", arguments: ["-rf", "Sources\(ps)Generated\(ps)FishyJoes.generated.swift"]).run()
-            try cmd("rm", arguments: ["-rf"] + sourceLocations).run()
+            try cmd("rm", "-rf", "bindings/swift-interfaces/generated").run()
```
mstokercricut, line 187: Should we collect these directories into a "generatedLocations" or "outputLocations" type thing then clear them in a loop? Might be useful to have a list of these directories for other purposes too...

Sources/FishyJoesExecute/CodeGen.swift
```diff
@@ -160,42 +168,180 @@ extension CodeGen {
...
+                config.requiredModules.flatMap {
+                    ["\n"] +
+                    Array(repeating: "    ", count: tabs) +
+                    [".product(name: \"\($0)-\(name)\", package: \"\($0)-bindings\"),"]
+                }.joined()
+            }
+
+            try cmd("cat")
+                .output(overwritingFile: "\(swiftBindingsRoot)/Package.swift")
+                .input(
```
mstokercricut, line 214: This is a pretty long string. Should it be turned into a template-type thing? I guess it's pretty key to understanding the rest of what happens afterward in this file...

mstokercricut (approved):
All comments are nits

Sources/FishyJoesExecute/DockerContext.swift
cobbal: Yes. It deletes it as soon as it can, and passing credentials this way is opt-in with a command-line switch.

Hopefully it's no worse than npm's password storage solution.

Sources/FishyJoesExecute/Platform.swift
cobbal: Yeah, this is probably wrong, but node is completely broken in this PR anyway.

Sources/FishyJoesExecute/CodeGen.swift
cobbal: It has a lot of customization in it, and the current templating system is just a find-and-replace. We may want a more powerful templating system at some point

PR #218, opened 2024-05-23
Add Default Computed Properties support to exported Protocols

dvosscricut:
Add Default Computed Properties support to exported Protocols

https://cricut.visualstudio.com/Cricut/_workitems/edit/476436


Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
```diff
@@ -284,12 +284,56 @@ struct TranslatedProtocol: TranslatedType {
...
+                                methodParamsStr.append("\(param.name)")
+                            }
+                        }
+                        fragment.output("\(method.isThrowing ? "try " : "")\(method.isAsync ? "await ": "")wrapped.\(method.callName)(\(methodParamsStr.joined(separator: ", ")))")
+                    }
+                }
+            }
+            fragment.output()
+        }
+
```
cobbal: This is a lot of duplicated code. The sans-method-struct and the sans-field-struct should be almost the same

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
```diff
@@ -836,8 +908,29 @@ struct TranslatedProtocol: TranslatedType {
...
+                        isStatic: true,
+                        isSuspend: false,
+                        isOverride: false,
+                        isDefaultImplementation: true,
+                        name: kotlinVariable.name,
+                        parameters: [(nil, name: "self", type: .named(package: nil, name: kotlinName), nil)],
+                        compatibilityOrder: [],
+                        returnType: kotlinVariable.type
+                    )
+                    kotlinVariable.body = "\tget() = __jni__default_\(defaultMethodForVariable.name)(this)"
```
cobbal: Tabs should not be used, especially when mixed with spaces

Sources/FishyJoesCore/Field.swift
```diff
@@ -20,7 +20,7 @@ struct Field: Hashable {
     let deprecation: Deprecation?
     let isDefaultImplementation: Bool
 
-    init?(_ variable: SourceryVariable, type: Type?) {
+    init?(_ variable: SourceryVariable, type: Type?, isDefaultImplementation: Bool) {
```
cobbal: What are the reasons for the differences between this and Method?

Sources/FishyJoesCore/TranslatedTypes/TranslatedStruct.swift
```diff
@@ -139,10 +148,15 @@ struct TranslatedStruct: TranslatedType {
...
-                        hasProperties ||= context.nodeTranslator.outputProperties(methods: defaultMethods, context: context, fragment: fragment, converterName: sourceType.name)
-                        hasProperties ||= context.nodeTranslator.outputProperties(computedVariables: computedVariables, context: context, fragment: fragment)
+                        hasProperties ||= context.nodeTranslator.outputProperties(methods: normalMethods, context: context, fragment: fragment, converterName: converterType.name)
+                        hasProperties ||= context.nodeTranslator.outputProperties(methods: defaultMethods, context: context, fragment: fragment, converterName: converterType.name)
+                        hasProperties ||= context.nodeTranslator.outputProperties(computedVariables: normalProperties, context: context, fragment: fragment, converterName: converterType.name)
+                        hasProperties ||= context.nodeTranslator.outputProperties(computedVariables: defaultProperties, context: context, fragment: fragment, converterName: converterType.name)
```
cobbal: splitting things out only to immediately recombine them

Sources/FishyJoesCore/TranslatedTypes/TranslatedStruct.swift
```diff
@@ -33,15 +33,24 @@ struct TranslatedStruct: TranslatedType {
...
         self.conformances = exportAnnotation.conformances
+
+        enforceMustHaveProperties()
+    }
+
+    func enforceMustHaveProperties() {
+        // https://kotlinlang.org/docs/data-classes.html
+        // * The primary constructor must have at least one parameter.
+        if storedVariables.isEmpty {
+            fatalError("☠️ Error on \(sourceType.name): Exported structs must have at least one stored property, it's the law 👮!")
```
cobbal, line 52: This is probably too strict of a restriction to place on structs. A struct without any members is perfectly valid, and we should be able to export it.

cobbal (changes requested):
I got part way through reviewing this, and was trying to test some changes, but the big question I just can't get past is why are default variables treated so differently from default methods? Is there a technical reason? I know you said something about definedInType, but I think that may be a symptom, not a cause.

The difference either needs to go away, or it needs to become obvious why it's there. Then it will be easier to review this without starting from scratch.

dvosscricut:
> I got part way through reviewing this, and was trying to test some changes, but the big question I just can't get past is why are default variables treated so differently from default methods? Is there a technical reason? I know you said something about definedInType, but I think that may be a symptom, not a cause.
> 
> The difference either needs to go away, or it needs to become obvious why it's there. Then it will be easier to review this without starting from scratch.

definedInType for default implementations of computed variables can be SourceryRuntime.Type instead of SourceryProtocol.

That's the reason for the difference.

I don't know if methods have the same problem, but I should check.

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
dvosscricut: Code duplication indeed! I have combined them now!

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
dvosscricut: spaces it is! 
space_sloth

Sources/FishyJoesCore/TranslatedTypes/TranslatedStruct.swift
dvosscricut: A data class in Kotlin (which is what KotlinProductClass exports to) without any members is invalid:

https://kotlinlang.org/docs/data-classes.html
* The primary constructor must have at least one parameter.

Sources/FishyJoesCore/TranslatedTypes/TranslatedStruct.swift
dvosscricut: good point. I think at one point they were different but then I fixed it but never recombined them.

dvosscricut: fixed now

Sources/FishyJoesCore/Field.swift
dvosscricut: It's needed due to definedInType for default implementations of computed variables can be SourceryRuntime.Type instead of SourceryProtocol. Which makes me have to pass in isDefaultImplementation explicitly instead of being able to use variable.definedInType is SourceryProtocol.

I'll check if methods have a similar problem, they may need the same treatment.

dvosscricut: methods have a similar problem of a default method definedInType not being a SourceryProtocol but instead being a SourceryRuntime.Type instead. I checked this out with a breakpoint on line 171 of Method.swift, and doing a `po prot.defaultMethods()[0].definedInType is SourceryProtocol` and it came up false.

It's a good point worthy of investigation. Why aren't they the same? Could they be handled the same?

dvosscricut: so it looks like this logic here in methodsPreferring would fail for methods that are not SourceryProtocol but ought to be `let isMostlyEqualMethodADefaultImplementation = mostlyEqualMethod.isExtension && (mostlyEqualMethod.definedInType is SourceryProtocol)`

So I should convert over the methods stuff to do it similarly to how the variables do it, with the isDefinedInProtocol property on the wrapper SourceryVariablePlus struct. And, I will!

dvosscricut: Ok, I've converted over SourceryMethod to handle methods like variables from Sourcery

Because method.definedInType is SourceryProtocol can be false even when it should be true, therefore we must keep track of isDefinedInProtocol instead to give us the that information that we need to compute isDefaultImplementation.

dvosscricut:
> > I got part way through reviewing this, and was trying to test some changes, but the big question I just can't get past is why are default variables treated so differently from default methods? Is there a technical reason? I know you said something about definedInType, but I think that may be a symptom, not a cause.
> > The difference either needs to go away, or it needs to become obvious why it's there. Then it will be easier to review this without starting from scratch.
> 
> definedInType for default implementations of computed variables can be SourceryRuntime.Type instead of SourceryProtocol.
> 
> That's the reason for the difference.
> 
> I don't know if methods have the same problem, but I should check.

@cobbal okay I checked and methods have the same problem, and I just did them the same as variables for the same solution.

Sources/FishyJoesCore/Field.swift
dvosscricut: OK, now I've gotten rid of SourceryMethodPlus and SourceryVariablePlus and just use Method.methods(type:) and Field.fields(type:). hooray!

Sources/FishyJoesCore/Field.swift
```diff
@@ -33,10 +33,97 @@ struct Field: Hashable {
         self.isPubliclyWritable =
             variable.isMutable && variable.accessLevel.write == .public
         self.isComputed = variable.isComputed
-        self.isDefaultImplementation =
-            type is SourceryProtocol && variable.definedInType?.isExtension == true
+        self.isDefaultImplementation = isDefaultImplementation
```
cobbal: This change can probably be reverted now.

cobbal: Sanity check, I dropped a `precondition(isDefaultImplementation == (type is SourceryProtocol && variable.definedInType?.isExtension == true))` and everything still worked. It should be safe to delete the `isDefaultImplementation` parameter to centralize the logic to this method.

Sources/FishyJoesCore/Method.swift
```diff
@@ -27,11 +27,7 @@ struct Method: Hashable {
     }
     let sourceKind: SourceKind
 
-    init?(_ method: SourceryMethod, type: Type?) {
-        self.init(method, type: type, protocolName: nil)
-    }
-
-    init?(_ method: SourceryMethod, type: Type?, protocolName: String?) {
+    init?(_ method: SourceryMethod, type: Type?, isDefaultImplementation: Bool, protocolName: String?) {
```
cobbal: This parameter too

cobbal (approved):
This is looking really good. Diff made a lot more sense to me this time around. One minor cleanup could still be done.

Sources/FishyJoesCore/Field.swift
dvosscricut: ok

Sources/FishyJoesCore/Method.swift
dvosscricut: <jimmy McLemming> "Okay!"

PR #220, opened 2024-05-29
AB#478509 Use BetterType instead of String for Conformances so we have a namespace for conformances

dvosscricut:
AB#478509 Use BetterType instead of String for Conformances so we have a namespace for conformances


Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
```diff
@@ -886,8 +888,8 @@ struct TranslatedProtocol: TranslatedType {
                 constructor: .reference,
                 fields: externalWitnessFields,
                 methods: externalWitnessMethods,
-                conformances: ["com.cricut.fishyjoes.runtime.SwiftReference(_swiftReference)"]
-            ).conforming(to: [sourceType.name], context: context)
+                conformances: [BetterType.named(.init(name: "SwiftReference(_swiftReference)", module: "fishyjoes.runtime"))]
```
cobbal: This is the name of a kotlin class, It should be either a `KType` or a string, BetterType refers to swift types. (Maybe it should be called SwiftType... I'll have to think about that)

Sources/FishyJoesCore/Unparse/CSharpClass.swift
```diff
@@ -42,22 +42,22 @@ class CSharpClass: NestedClass {
     var innerClasses: [CSharpClass] = []
     let fields: [Variable]
     let methods: [Method]
-    let conformances: [String]
+    let conformances: Set<BetterType>
```
cobbal: Same here

cobbal (changes requested):
Nice. That worked out well. There are just a couple places where a swift type is placed into a context that should have a different-lannguage type.

Sources/FishyJoesCore/TranslatedTypes/TranslatedProtocol.swift
dvosscricut: updated to use KotlinClass.KType inside The Kotlin Kingdom

We still use BetterType/SType when dwelling within the borders of the Swift Sovereign State.

Sources/FishyJoesCore/Unparse/CSharpClass.swift
dvosscricut: CSharpClass.CSType is now used within the C# Kingdom

Sources/FishyJoesCore/TranslatedTypes/TranslatedEnum+cSharp.swift
```diff
@@ -108,6 +108,10 @@ extension TranslatedEnum {
 
         let (enumFields, enumMethods) = CSharpClass.separate(fieldsAndMethods: fieldsAndMethods)
 
+        let csTypeConformances = conformances.map {
+            CSharpClass.CSType.named(package: $0.module, name: $0.nonNamespacedName)
```
cobbal: This should be resolved and the csharptype pulled from the translatedtype. The swift type may not be named the same thing.

cobbal (changes requested):

Sources/FishyJoesCore/TranslatedTypes/TranslatedEnum+cSharp.swift
dvosscricut: ok! <high pitched lemming voice>

Sources/FishyJoesCore/TranslatedTypes/TranslatedEnum+cSharp.swift
```diff
@@ -109,7 +109,8 @@ extension TranslatedEnum {
         let (enumFields, enumMethods) = CSharpClass.separate(fieldsAndMethods: fieldsAndMethods)
 
         let csTypeConformances = conformances.map {
-            CSharpClass.CSType.named(package: $0.module, name: $0.nonNamespacedName)
+            let resolved = context.resolve(type: $0)
+            return CSharpClass.CSType.named(package: resolved.sourceType.module, name: resolved.sourceType.nonNamespacedName)
```
cobbal: Use this: https://github.com/cricut/FishyJoes/blob/main/Sources/FishyJoesCore/TranslatedTypes/TranslatedType.swift#L11

dvosscricut: ok, I did it for kotlinType as well

cobbal (approved):
Looks very good

PR #222, opened 2024-06-04
AB#478507 Add Tests and Support for EmptyEnum/Struct/Class

dvosscricut:
Add Tests and Support for EmptyEnum/Struct/Class

EmptyEnum worked fine already but Struct and Class needed equals/hashCode and toString methods added since they can no longer be auto handled by Kotlin data class since that requires at least one stored parameter so we must go with a general Kotlin class which does not have those methods autohandled so we must implement them ourselves.

https://cricut.visualstudio.com/Cricut/_workitems/edit/478507


Sources/FishyJoesCore/Unparse/KotlinProductClass.swift
```diff
@@ -76,6 +90,58 @@ class KotlinProductClass: KotlinClass {
 
             fragment.blankLine()
 
+            if generateEqualsMethod {
+                // generate equals and hashCode functions
+                fragment.outputBlock("override fun equals(other: Any?): kotlin.Boolean {") {
```
cobbal: Unparse is the wrong place to do this. They should be created in TranslatedStruct for empty structs like they're created in TranslatedReference for equatable references.

cobbal (changes requested):

Sources/FishyJoesCore/Unparse/KotlinProductClass.swift
dvosscricut: "ok!"

Sources/FishyJoesCore/TranslatedTypes/TranslatedReference.swift
```diff
@@ -271,6 +271,112 @@ struct TranslatedReference: TranslatedType {
...
+                        parameters: [],
+                        compatibilityOrder: [],
+                        returnType: .named(package: nil, name: "kotlin.String"),
+                        deprecation: nil,
+                        body: "return \"\(kotlinName)(\(fields.map { "\($0.name) = ${\($0.name)}" }.joined(separator: ", ")))\"",
+                        isBodyInline: true
+                    )
+                )
+            )
+        }
```
cobbal: Why did TranslatedReference need to change? Was there something wrong with the way it was already handling equatable?

Sources/FishyJoesCore/TranslatedTypes/TranslatedStruct.swift
```diff
@@ -291,14 +281,127 @@ struct TranslatedStruct: TranslatedType {
...
+            var bodyBuilder = [String]()
+
+            bodyBuilder.append("if (this === other) {")
+            bodyBuilder.append("    return true")
+            bodyBuilder.append("}")
+            bodyBuilder.append("if (other !is \(kotlinName)) {")
+            bodyBuilder.append("    return false")
+            bodyBuilder.append("}")
+
+            let fieldsChecks = fields.map {
```
cobbal: If there are no stored fields, this isn't necessary and probably shouldn't be done. Here's an example of a workaround for the restriction: https://discuss.kotlinlang.org/t/data-class-without-arguments-deprecated-in-1-0-why/60/15

Sources/FishyJoesCore/Unparse/KotlinProductClass.swift
```diff
@@ -37,21 +37,23 @@ class KotlinProductClass: KotlinClass {
...
         switch constructor {
         case .reference:
             fragment.output("\(isPrivate ? "private " : "")class \(unqualifiedName) private constructor(_swiftReference: Long)", newLineTerminated: false)
-        case .`public`(let fields, let arguments):
+        case .`public`(let storedVariables, let arguments):
             var classDeclaration: String
-            if isPrivate {
+
+            if isPrivate || storedVariables.isEmpty {
                 classDeclaration = "class \(unqualifiedName) private constructor"
```
cobbal, line 48: The constructor should be public for structs

cobbal (changes requested):

Sources/FishyJoesCore/TranslatedTypes/TranslatedReference.swift
dvosscricut: because I didn't understand that swift class must conform to Hashable in order to get those __jni_swiftEquals type functions to autogenerate. But now, thanks to you, I know!

So, you're right TranslatedReference doesn't have to change!

Sources/FishyJoesCore/TranslatedTypes/TranslatedStruct.swift
dvosscricut: ok

Sources/FishyJoesCore/Unparse/KotlinProductClass.swift
dvosscricut: ok

cobbal (approved):

PR #224, opened 2024-06-08
Dvoss/create double doubley

dvosscricut:
because calling napi_create_double twice works on windows, just once won't do, for the first time there's some bug somewhere that clobbers the returned value from it.

### DIFF
--- a/integration-tests/TestAPI-bindings/node-test/Protocols.test.ts
+++ b/integration-tests/TestAPI-bindings/node-test/Protocols.test.ts
@@ -165,27 +165,32 @@ test('testProtocolStruct', () => {
 
 test('testProtocolClass', () => {
     const testProtocolClass = TestAPI.TestProtocolClass.init("Step inside it's a wilder ride!");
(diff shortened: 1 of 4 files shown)


integration-tests/TestAPI-bindings/node-test/Protocols.test.ts
```diff
@@ -165,6 +165,11 @@ test('testProtocolStruct', () => {
 
 test('testProtocolClass', () => {
     const testProtocolClass = TestAPI.TestProtocolClass.init("Step inside it's a wilder ride!");
+
+    // for (let i=0; i < 10; i++) {
+    //     var w = testProtocolClass.wombat(undefined)
+    //     console.log("wombat: " + w);
+    // }
```
cobbal, line 172: Is this helpful to commit?

cobbal (approved):

integration-tests/TestAPI-bindings/node-test/Protocols.test.ts
dvosscricut: for future testing, yes, I thought it would be helpful

PR #229, opened 2024-06-11
AB#492335 Filter conformances to FishyJoes exported ones

dvosscricut:
Filter conformances to FishyJoes exported ones

To fix CriGeo-bindings problem

It appears that Sourcery does not put in type.implements any protocols that are declared in the initial sttuct/class/enum declaration if they are not exported to FishyJoes, but it DOES put in type.implements any protocols that are declared in extensions to the struct/class/enum.

To deal with this, we must find first find all fishyjoes exported protocols, then filter type.implements by them in the translated struct/class/enum initializers when assigning to conformances.

https://cricut.visualstudio.com/Cricut/_workitems/edit/492335


cobbal (changes requested):
I think this could be done more cleanly by adding a `func exportedConformances(in context: FishyJoesContext) -> [TranslatedProtocol]` to the TranslatedType extension.

Due to the nature of cycles, TranslatedTypes avoid needing to know about other TranslatedTypes while being initialized, and wait until the fragment generation stages to resolve the BetterTypes into TranslatedTypes.

Similarly, the decision about which protocols types should be exported should be delayed until conformances are used in the fragment generation. The suggested `exportedConformances` method should make this fairly low-repeated-code.

Sources/FishyJoesCore/TranslatedTypes/TranslatedType.swift
```diff
@@ -91,6 +91,19 @@ extension TranslatedType {
...
+extension TranslatedType {
+    func exportedConformances(in context: FishyJoesContext) -> [TranslatedProtocol] {
+        let implementedExportedProtocols = conformances.filter { conformance in
+            !context.exportedProtocolSwiftTypes.map { $0.name }.filter { $0.contains(conformance.name)}.isEmpty
+        }
+        let exportedConformances = implementedExportedProtocols.compactMap {
+            let resolved = try? context.resolve(type: $0)
+            return resolved as? TranslatedProtocol
+        }
+        return exportedConformances
```
cobbal: This could be much simpler. The exportedProtocolSwiftTypes can be entirely deleted.
```suggestion
        conformances.compactMap { try? context.tryResolve(type: $0) as? TranslatedProtocol }
```

cobbal (changes requested):

dvosscricut:
oops closed by accident

cobbal (approved):

Sources/FishyJoesCore/TranslatedTypes/TranslatedType.swift
dvosscricut: good point

PR #230, opened 2024-06-12
AB#492339 Combine Generated Source files to work around Windows swift file source argument count limit

dvosscricut:
Combine Generated Source files to work around Windows swift file source argument count limit

Gotten from Konrad Rodzik's https://github.com/cricut/Tesseract-bindings/blob/main/combine_generated_sources.sh

https://cricut.visualstudio.com/Cricut/_workitems/edit/492339


dvosscricut:
After discussion with Matt and Andrew, this should be done in FishyJoes, not a shell script. So, closing this PR.

PR #236, opened 2024-06-21
Fix hashCode bug on c# (wrong name in c# code). Fix eq bug on C# if other is different type than self.

dvosscricut:
Fix hashCode bug on c# (wrong name in c# code). Fix eq bug on C# if other is different type than self.


Sources/FishyJoesCore/BetterType.swift
```diff
@@ -150,14 +150,12 @@ extension BetterType {
             self = .named(.foundation(parts.dropFirst().joined(separator: ".")))
         } else if
             parts.count > 1,
-            case let .named(baseName) = try? context.tryResolve(type: .named(.init(name: parts[0], module: nil))).sourceType
-        {
+            case let .named(baseName) = try? context.tryResolve(type: .named(.init(name: parts[0], module: nil))).sourceType {
```
cobbal, line 153: This change only makes things harder to read. Same with the other indentation-only changes.

cobbal (approved):
The code looks fine, but the indentation changes put the if condition at the same level as the block body with no visual separation.

PR #237, opened 2024-06-24
Revert indentation changes

cobbal:


dvosscricut (changes requested):
This is what latest SwiftLint version does 0.55.1

cobbal:
> This is what latest SwiftLint version does 0.55.1

Yes. This is a regression in swiftlint and is being tracked upstream: https://github.com/realm/SwiftLint/issues/5602

The repo uses a pinned version of swiftlint (0.53.0 currently) so we don't get auto-upgraded into new lints before we evaluate them. 

This is just a much harder style of whitespace to read, and it was deliberately indented the way it was.

cobbal:
To run a particular version of swiftlint on your local machine, the [mint](https://github.com/yonaskolb/Mint) package manager makes it easy: `mint run swiftlint@0.53.0 --quiet --strict`

dvosscricut (approved):
ok I'll just use mint and run SwiftLint 0.53.1 until we decide to update SwiftLint version on  CI then

PR #238, opened 2024-06-25
one repo node

cobbal:


.github/workflows/integration-tests.yaml
```diff
@@ -57,13 +57,13 @@ jobs:
         run: |
           set -euxo pipefail
 
-          cd integration-tests/TestAPI-bindings
+          cd integration-tests/TestAPI
 
-          swift run -- fishy-joes generate
+          FISHYJOES=1 swift run -- fishy-joes generate
```
mstokercricut, line 62: Sad that we need this, but ultimately I think it will simplify Swift build workflows that use the same source files. Its a good addition.

mstokercricut (approved):

PR #239, opened 2024-06-25
refactor codegen into phases grouped by language

cobbal:


mstokercricut (approved):
I love it all except the nit: "phases" directory should be called "Phases".

PR #241, opened 2024-06-26
Fix toString for Dart SwiftReferences. Add tests. Add toString to Node objects.

dvosscricut:
Fix toString for Dart SwiftReferences. Add tests. Add toString to Node objects.

If an export is a SwiftReference, then we should use the runtime's toString method to get the swift objects string description.

I'm not sure if we should generate toString methods on the foreign language side for non reference exports for all languages, but I'm doing it for dart.

The foreign language side built in toString is not symmetric across languages, each language that has it built in will do it differently.

### DIFF
--- a/Sources/FishyJoesCore/TranslatedTypes/TranslatedReference.swift
+++ b/Sources/FishyJoesCore/TranslatedTypes/TranslatedReference.swift
@@ -133,6 +133,7 @@ struct TranslatedReference: TranslatedType {
 
             fragment.output("@available(*, deprecated, message: \"Not actually deprecated, but this silences warnings because it may refer to deprecated methods\")")
             fragment.outputBlock("public static func nodeSetup(env: NAPI.Env, module: NAPI.Value) throws {") {
+                // fragment.output("print(\"setting up \(sourceType.name)\")")
                 fragment.outputBlock("let nodeClass = try NodeClass(") {
                     fragment.output("env: env,")
                     fragment.output("name: \"\(nodeName)\",")
@@ -144,7 +145,7 @@ struct TranslatedReference: TranslatedType {
                             fragment.outputBlock(".method { env, info in", closeWith: "},") {
                                 fragment.outputBlock(#"FishyJoesNodeRuntime.callbackBody(env, info, name: "toString", expectedArgumentCount: 0, hasNamedOptions: false) { env in"#, closeWith: "}") {
                                     fragment.outputBlock("let result = try Swift.String.toNode(") {
-                                        fragment.output("\"\\(env.this(converter: \(sourceType.name).self))\",")
+                                        fragment.output("\"\\(env.this(converter: \(converterType.name).self))\",")
                                         fragment.output("env: env.env")
                                     }
                                     fragment.output("return result")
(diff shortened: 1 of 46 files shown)


Sources/FishyJoesCore/TranslatedTypes/TranslatedReference.swift
```diff
@@ -133,17 +133,24 @@ struct TranslatedReference: TranslatedType {
 
             fragment.output("@available(*, deprecated, message: \"Not actually deprecated, but this silences warnings because it may refer to deprecated methods\")")
             fragment.outputBlock("public static func nodeSetup(env: NAPI.Env, module: NAPI.Value) throws {") {
-                // fragment.output("print(\"setting up \(sourceType.name)\")")
```
cobbal, line 136: This is a useful debugging aid when diagnosing library-load crashes. Please leave it in.

Sources/FishyJoesCore/TranslatedTypes/TranslatedReference.swift
```diff
@@ -133,17 +133,24 @@ struct TranslatedReference: TranslatedType {
...
                         var hasProperties = false
                         hasProperties ||= context.nodeTranslator.outputProperties(methods: methods, context: context, fragment: fragment, converterName: converterType.name)
                         hasProperties ||= context.nodeTranslator.outputProperties(computedVariables: computedVariables, context: context, fragment: fragment, converterName: converterType.name)
-                        if !hasProperties {
-                            fragment.output(":")
+                        fragment.outputBlock(#""toString": ("#) {
+                            fragment.outputBlock(".method { env, info in", closeWith: "},") {
+                                fragment.outputBlock(#"FishyJoesNodeRuntime.callbackBody(env, info, name: "toString", expectedArgumentCount: 0, hasNamedOptions: false) { env in"#, closeWith: "}") {
+                                    fragment.outputBlock("let result = try Swift.String.toNode(") {
+                                        fragment.output("\"\\(env.this(converter: \(sourceType.name).self))\",")
```
cobbal: nit: this should be converterType, not sourceType. Even though they're the same for TranslatedReference.

Sources/FishyJoesCore/Unparse/DartProductClass.swift
```diff
@@ -36,8 +36,14 @@ class DartProductClass: DartClass {
 
     private func toStringImpl(fields: [Variable], fragment: SourceFragment) {
         fragment.output("@override")
-        fragment.output("String toString() => '\(unqualifiedName)(", newLineTerminated: false)
-        let toStringParamsString = fields.map { "\(DartClass.deforbidify($0.name)): $\(DartClass.deforbidify($0.name))" }.joined(separator: ", ")
+        fragment.output("String toString() => '\(module.name).\(unqualifiedName)(", newLineTerminated: false)
+
+        let toStringParamsString = fields.filter {
+            $0.name != "hashCode"
```
cobbal: This should not be needed. If hashCode is showing up as a stored field, something has gone wrong prior to this.

cobbal: This may have been fixed already by the line 81 change, in which case the filter should be removed

cobbal (changes requested):

Sources/FishyJoesCore/TranslatedTypes/TranslatedReference.swift
dvosscricut: ok

Sources/FishyJoesCore/TranslatedTypes/TranslatedReference.swift
dvosscricut: hence the argument name "converter: " ha ha.

Sources/FishyJoesCore/Unparse/DartProductClass.swift
dvosscricut: yes, it's fixed by line 81 change, no more need for filter

Sources/FishyJoesCore/TranslatedTypes/TranslatedReference.swift
dvosscricut: fixed

cobbal (approved):

PR #242, opened 2024-06-27
TODONE: Type check node types in fromNode and constructor

dvosscricut:
Type check node types in fromNode and constructor

Make the todo todone


dvosscricut:
This TODO is larger in scope than we want todo, as in type checking napi arguments correctly instead of just a subset of them correctly.  As in an optional int, if nil, will be napi_undefined, but an optional int, if not nil, say '2', will be napi_number.

@cobbal has given permission to just simply remove this TODO, as after his consideration he concluded that we will never want todo this in the future.

PR #252, opened 2024-08-19
Switch to reactor model for wasm

cobbal:
C++ was doing some deinitialization that finally broke the command model

### DIFF
--- a/Sources/FishyJoesNodeRuntime/Templates/wasm-napi.js
+++ b/Sources/FishyJoesNodeRuntime/Templates/wasm-napi.js
@@ -1899,6 +1899,7 @@ export class NAPI {
     this.indirectFunctionTable = instance.exports.__indirect_function_table;
 
     this.wasi.start(instance);
+    instance.exports._initialize?.();
 
     this.errorMessageTable = this.allocateErrorMessages();
     this.extendedErrorInfoPtr = this.malloc(16);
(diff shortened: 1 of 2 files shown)


Sources/FishyJoesNodeRuntime/Templates/wasm-napi.js
```diff
@@ -1899,6 +1899,7 @@ export class NAPI {
     this.indirectFunctionTable = instance.exports.__indirect_function_table;
 
     this.wasi.start(instance);
+    instance.exports._initialize?.();
```
ejohn-cricut, line 1902: Will you get an error calling start? I was under the impression it was one or the other, depending on the model used

ejohn-cricut (approved):

Sources/FishyJoesNodeRuntime/Templates/wasm-napi.js
cobbal: The wasi.start function contains these lines:
```
        if (exports._start) {
            exports._start();
        }
```
so it's safe

ejohn-cricut: There may be a counterpart to `this.wasi.start(instance);` something like `this.wasi.initialize(instance);`

PR #258, opened 2024-10-17
Add `extraDynamicLibraries` parameter to fishy-joes.yaml 

dvosscricut:
This is to support CanvasModel-Swift being depended upon by both CriSVG-bindings and CriCanvas-bindings now.

So we must change from CanvasModel-Swift being linked statically into both -bindings repos, which results in duplicate symbols warnings/errors, to it being linked dynamically. Since CanvasModel-Swift is not a -bindings repo, it needs a separate input argument in fishy-joes.yaml

### DIFF
--- a/Sources/FishyJoesExecute/FishyJoesConfig.swift
+++ b/Sources/FishyJoesExecute/FishyJoesConfig.swift
@@ -6,6 +6,7 @@ struct FishyJoesConfig: Codable {
     let module: String
     let publishRepository: String?
     let requiredModules: [String]
+    let extraDynamicLibraries: [String]
     let excludeSources: [String]
 
     static func readFromFile() throws -> FishyJoesConfig {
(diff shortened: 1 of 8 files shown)


Sources/FishyJoesExecute/FishyJoesConfig.swift
```diff
@@ -43,12 +44,18 @@ struct FishyJoesConfig: Codable {
             }
             return list
         }
+        let extraDynamicLibraries = try configDictionary["extraDynamicLibraries"].map { obj -> [String] in
+            guard let list = obj as? [String] else {
+                throw ValidationError("fishy-joes.yaml value for key `extraDynamicLibraries` is not an array of file or directory paths")
```
cobbal: This error message is probably copy-pasted. I don't think directories are allowed, and I think it expects file names, not paths.

cobbal (approved):
Looks good

Sources/FishyJoesExecute/FishyJoesConfig.swift
dvosscricut: ok I'll fix the message

PR #260, opened 2024-10-18
fix: fix packing to handle xdls correctly

dvosscricut:
move tar of dart libs to pack from -bindings.yaml to pack command in CodeGen
Add xdls to dart cp commands in pack
remove dependency's xdls from c-sharp common lib directory before pack (keep direct xdls)

### DIFF
--- a/Sources/FishyJoesExecute/CodeGen.swift
+++ b/Sources/FishyJoesExecute/CodeGen.swift
@@ -694,7 +694,8 @@ extension CodeGen {
                         .run()
                 case .kotlinSystem, .kotlinAndroid:
                     // Install the module library and interfacing JNI library
-                    try extraDynamicLibsToInstall.forEach {
+                    // Don't copy over any dependency's extraDynamicLibraries for Kotlin because it's bundled by maven from the dependency.
+                    try config.extraDynamicLibraries.forEach {
                         try installLibrary($0)
                     }
                     try installLibrary(config.module)
--- a/Sources/FishyJoesExecute/CodeGen.swift
+++ b/Sources/FishyJoesExecute/CodeGen.swift
@@ -621,15 +621,6 @@ extension CodeGen {
                     for dependency in nodeDependencies {
                         npmDependencies["@cricut/\(dependency.npmPackageName)"] = dependency.npmModuleVersion
                     }
-                    for (xdl, xdlRepoName) in zip(config.extraDynamicLibraries, config.extraDynamicLibrariesRepoNames) {
-                        guard let dependency = packageInfo.dependencyMap[xdlRepoName] else {
-                            fatalError("Could not locate dependency \(xdlRepoName) in Package.swift")
-                        }
-                        let dependencyNpmPackageName = "\(config.module.lowercased())-\(platform.nodeExecutionEnvironment)"
-                        // Use a file-local dependency
-                        let dependencyNpmModuleVersion = "file:\(dependency.localPath)\(ps)output\(ps)\(platform.platform)"
-                        npmDependencies["@cricut/\(dependencyNpmPackageName)"] = dependencyNpmModuleVersion
-                    }
                     var package = NPMPackage(
                         config: config,
                         platform: platform,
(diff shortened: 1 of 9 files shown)


Sources/FishyJoesExecute/CodeGen.swift
```diff
@@ -152,6 +152,15 @@ extension CodeGen {
...
+            $0.key == config.module ? nil : "\($0.value)\(ps)..\(ps)"
+        }.map(FishyJoesConfig.readFromFile(basePath:))
+
+        // extraDynamicLibsToInstall is different from config.extraDynamicLibraries because it includes the extraDynamicLibraries of all the dependencies as well, instead of the ones just for this module. This is needed because say if CriSVG depends on the dynamic library CanvasModel and CriCanvas also depends on CanvasModel and on CriSVG as well, then we must install the CanvasModel library i.e. copy over the dynamic library to where the language environment needs it to run and test with, but we don't want to pass it to the generate/sourcery step where it's used to generate the kotlin LoaderRepresentative.kt file, because CriSVG will already load it there, so if we try to load the same dynamic CanvasModel library in CriCanvas's LoaderRepresentative we will get an error on Windows and Linux.
+        var extraDynamicLibsToInstall = Set<String>(config.extraDynamicLibraries)
+        fishyJoesYamlConfigs.forEach { extraDynamicLibsToInstall.formUnion($0.extraDynamicLibraries) }
```
cobbal: I worry that this will package the canvas model twice, once in the CriSVG-bindings release and once in the CriCanvas-bindings. It should only be included in the SVG package.

This might be pointing towards an inconsistency in how FishyJoes does dart packaging with how it does the rest of the languages.

Sources/FishyJoesExecute/CodeGen.swift
```diff
@@ -685,6 +694,7 @@ extension CodeGen {
                         .run()
                 case .kotlinSystem, .kotlinAndroid:
                     // Install the module library and interfacing JNI library
+                    // Don't copy over any dependency's extraDynamicLibraries for Kotlin because it's bundled by maven from the dependency.
```
dvosscricut, line 692: @cobbal  here is the change

Sources/FishyJoesExecute/CodeGen.swift
```diff
@@ -637,6 +642,13 @@ extension CodeGen {
                             package_directory=".."
                         fi
 
+                        echo "npm_package_version: ${npm_package_version}"
+                        echo "package_directory: ${package_directory}"
+                        echo "pwd: "
+                        pwd
+
+                        echo "finding: "
+                        find . -iname "*crisvg*"
```
cobbal: This probably shouldn't be committed to this repo

Sources/FishyJoesExecute/CodeGen.swift
```diff
@@ -859,7 +904,14 @@ extension CodeGen {
                         (path: "windows\(ps)CMakeLists.txt", required: true),
                         (path: "windows\(ps)native\(ps)\(config.module).dll", required: false),
                         (path: "windows\(ps)native\(ps)\(config.module)-iota.dll", required: false),
-                    ]
+                    ] +
+                    config.extraDynamicLibraries.flatMap {
+                        [
+                            (path: "macos\(ps)native\(ps)lib\($0).dylib", required: true),
+                            (path: "linux\(ps)native\(ps)lib\($0).so", required: true),
+                            (path: "windows\(ps)native\(ps)\($0).dll", required: true)
```
cobbal: These should not be marked as required, because they're platform specific. Linux likely won't have the windows library, etc.

cobbal (changes requested):
Looks pretty good. Just 2 minor things

Sources/FishyJoesExecute/CodeGen.swift
dvosscricut: oh ha ha. good catch

Sources/FishyJoesExecute/CodeGen.swift
dvosscricut: ok

cobbal (approved):

PR #263, opened 2024-11-19
simplify windows instructions

cobbal:


README.md
```diff
@@ -168,47 +168,38 @@ public struct Foo {
...
-https://www.swift.org/download/ click on "Older Releases" to find it
+0. Install VS with x64 tools:
+```powershell
+winget install --id Microsoft.VisualStudio.2022.Professional --exact --force --custom "--add Microsoft.VisualStudio.Component.Windows11SDK.22000 --add Microsoft.VisualStudio.Component.VC.Tools.x86.x64"
+```
 
-The instructions in https://www.swift.org/install/windows/traditional/ aren't right for us since we are using Windows SDK 10.0.22000.0 and MSVC build tools 14.36.3253, plus in order to actually have the vcvarsall.bat file exist you need "Desktop development with C++" installed, but...
+1. Install swift
+```powershell
+winget install --id Swift.Toolchain --exact --force --architecture x64
```
dvosscricut, line 178: does this install the latest version of Swift toolchain? 

Because we want Swift 5.10 or 5.10.1, not Swift 6.

README.md
```diff
@@ -168,47 +168,38 @@ public struct Foo {
...
-* Windows 11 SDK (10.0.22000.0)
-* MSVC v143 - VS 2022 C++ x64/x86 build tools (v14.36-17.6) (don't select `(Latest)` because Swift 5.10 needs clang 16 and `(Latest)` will required clang 17 in some of the c headers for the libraries)
-
-Install these
-
-### Install Swift 5.10 (or 5.10.1):
-https://www.swift.org/download/ click on "Older Releases" to find it
+0. Install VS with x64 tools:
+```powershell
+winget install --id Microsoft.VisualStudio.2022.Professional --exact --force --custom "--add Microsoft.VisualStudio.Component.Windows11SDK.22000 --add Microsoft.VisualStudio.Component.VC.Tools.x86.x64"
```
dvosscricut, line 173: does this install the latest build tools? Because if it does it won't work with Swift 5.10/5.10.1 because that uses clang 16 instead of 17

dvosscricut (changes requested):
BiFrost-bindings CI useswindows-2019 runner for Windows and [self-hosted, solo-macos] runners for Mac, which have Swift 5.10 on them.

So this setup will not replicate that version. It's okay to have a separate section "Swift 6" or "Swift latest", but the Swift 5.10 setup instructions should be retained under a separate section so that we can debug/repro problems with BiFrost with the same Swift version they are using.

Screenshot 2024-11-19 at 7 09 10 PM

Screenshot 2024-11-19 at 7 07 27 PM

curiously, ubuntu uses Swift 5.10.1 instead of 5.10 for BiFrost actions:
Screenshot 2024-11-19 at 7 12 13 PM

dvosscricut (approved):
I'll put the windows 5.10 instructions in docs/windows-swift-5.10.md on a separate PR

PR #266, opened 2024-11-22
fix kotlin sets

cobbal:


mstokercricut (approved):
I love that TODOs disappeared with this!

PR #267, opened 2024-12-17
workaround for duplicate constructor calls

cobbal:


mstokercricut (approved):
These changes all seem appropriate. Nice that we have less built-in types to register as well.

mstokercricut:
Merged as Andrew is on PTO.

PR #283, opened 2025-04-14
Bifrost customizations

cobbal:
Creates a better system for applying customizations to yaml and json files. More generation can now happen in the generation phase, and NodePhase and all can participate in templating to better organize things by language.


Sources/FishyJoesExecute/CodeGen.swift
```diff
@@ -217,6 +217,9 @@ extension CodeGen {
         // "bindings/swift-interfaces/generated/\(config.module)-bindings"
         let swiftBindingsRoot = buildConfig.packagePath
 
+        // Whenever any code generation takes place, generate for every language and platform
+        let generationPhases: [any Phases] = Platform.allCases.map(phases(for:))
```
mstokercricut, line 221: I'm glad this is getting stated explicitly. Someday we might want to generate for only specific languages as requested.

Breaking the generate step into smaller parts like what you've done for the build and test steps would also be an improvement.

Sources/FishyJoesExecute/PackageInit.swift
```diff
@@ -151,91 +217,15 @@ public struct PackageInit: ParsableCommand {
...
-                }
-                if let version = dependency.versionInNugetFormat {
-                    return #"<ItemGroup><PackageReference Include="\#($0.nuget)" Version="\#(version)" /></ItemGroup>"#
-                } else {
-                    let path = relativePath(of: dependency.localPath, relativeTo: "bindings/c-sharp/generated/Cricut.\(config.module)/")
-                    dependencyPaths.append("$(MSBuildThisFileDirectory)\(path)/\($0.nupkgsPath)")
-                    return #"<ItemGroup><PackageReference Include="\#($0.nuget)" Version="[0.0.1-unknown]" /></ItemGroup>"#
+        // Collect language-specific replacements from phases
+        for phases in phasesList ?? [] {
+            for (key, value) in try phases.generationPhaseTemplateReplacements() {
```
mstokercricut, line 222: Great improvement.

mstokercricut (approved):
Improvements were worth the extra time.

PR #288, opened 2025-04-24
Make hand-crafted template files not be marked as "generated"

cobbal:
- remove "generated" from checked-in template files
 - Apply autogen warning to all files marked as template


mstokercricut (approved):
These look like good changes to me.

PR #293, opened 2025-04-29
Extract FileTemplater component out of the inside of PackageInit CLI command

cobbal:
"PackageInit" was an increasingly bad name for a large part of the generation phase


cobbal:
Also, if you have a better name than "file templater" I'm all ears

mstokercricut (approved):

PR #296, opened 2025-05-01
Use swift-wasm 6.1, which involves using a separate toolchain/sdk combo

cobbal:
compiling for wasm now depends on having swiftly installed, which is the swift.org-recommended way to install swift


mstokercricut (approved):
That cleans up a lot of things. Sad build.gradle.kts doesn't read the tools version directly. Someday.

PR #297, opened 2025-05-09
external sourcery

cobbal:
This PR primarily remove Sourcery as a swift dependency from FishyJoes, instead adding a binary dependency on mint to install the needed version of sourcery. This also allows for projects to override the version of sourcery they are using (no guarantees if sourcery makes a major change).


Sources/SourceryDataModel/SourceryDataModel.swift
cobbal, line 1: This file is the new interface between sourcery types and fishyjoes types

mstokercricut (approved):
Not a lot to say here. A mostly straightforward move from SourceryRuntime to SourceryDataModel. Nice cleanup in lots of areas as well. The inhabitedCache was an interesting addition.

PR #301, opened 2025-05-27
Don't use Package.resolved for picking a version

cobbal:
It turns out, this causes many more reliability problems than it fixes. In large part because Package.resolved isn't committed, leading to non-reproducible builds if an upstream repo publishes a new release.

Better to have a little extra manual work.


mstokercricut (approved):
This is a good change for now, while we wait to do up-to-next-major for all build systems.

PR #302, opened 2025-06-02
Use URI instead of URL in java

cobbal:
It more closely matches Swift's URL. There still are some differences. Hopefully they don't show up.


Sources/FishyJoesNodeRuntime/Templates/wasm-napi.js
```diff
@@ -1438,15 +1438,6 @@ export class NAPI {
...
       }),
-      napi_create_reference: this.wrap((envPtr, valueIdx, initialRefcount, resultPtr) => {
-        const value = this.load(valueIdx);
-        this.writeU32(resultPtr, this.createReference(value, initialRefcount));
-        return NAPI_OK;
-      }),
-      napi_delete_reference: (envPtr, refIdx) => {
-        delete this.references[refIdx];
-        return NAPI_OK;
-      },
```
cobbal, line 1449: This deletion is unrelated to the rest of the PR, but these functions are exact duplicates of other ones.

integration-tests/TestAPI/Sources/TestAPI/URLs.swift
```diff
@@ -16,6 +16,32 @@ public enum URLs {
     /// <!-- FishyJoes.export(localFile) -->
     public static let localFile: URL = { URL(string: "file:///Users/someuser/Desktop/a.png")! }()
 
+    /// <!-- FishyJoes.export(dataImage) -->
+    public static let dataImage: URL = {
```
mstokercricut, line 20: Glad to have this in here

mstokercricut (approved):

PR #306, opened 2025-06-26
Document enum cases consistently across languages and enum types

cobbal:


mstokercricut (approved):
Looks reasonable for enum case docs. Nice to clean up the "neutral" code too.

PR #308, opened 2025-09-18
Run prebuild on same shell as build

cschear:
Fixes errors when running git commands on linux.

image


Sources/FishyJoesExecute/Resources/bindings-template/workflows/__TEMPLATE_CAPS__-c-sharp.yaml
```diff
@@ -59,7 +59,7 @@ jobs:
           scope: '@cricut'
 
       - name: Pre-build hook
-        shell: bash
```
cobbal, line 62: Sadly we can't do this because zsh has no windows port. Instead we use bash for cross-platform-ness

cobbal (changes requested):
The git config looks like a good change though

PR #313, opened 2025-10-12
add optional config for generating workflow files

cschear:
Add an optional config param that allows us to skip the github workflow generation. For CriAnalytics this will allow the code generation to happen as part of the build - allowing folks to make changes to the event definitions only.

https://github.com/cricut/CriAnalytics/tree/main/Sources/EventDefinitions


cobbal:
I think we discussed this at the time and decided it wasn't needed. If I'm remembering wrong, feel free to re-open.

PR #314, opened 2025-10-15
make regenerate-check attach a downloadable patch

cobbal:
If the regenerate-check workflow fails, it will now generate and attach a patch file to the workflow with instructions on how to apply it. This provides an alternative for people who can't run fishy-joes generate and still need to develop.

Example failure: https://github.com/cricut/CriGeo/actions/runs/18538689082/job/52840389368


salminalex (approved):

mstokercricut (approved):
This is a great idea!!! People can at least apply the patch if they are on Windows and the patch is simple.

PR #330, opened 2026-03-02
Fishyjoes 7 prerelease

serhiiyaremychecw:
Fishyjoes 7 prerelease


mstokercricut (approved):
This all looks reasonable to me. I can't really comment on the JavaScript event loop code, though. Probably should run that by Andrew.

PR #336, opened 2026-04-16
Dart 3.9+ tag_pattern

serhiiyaremychecw:
## Summary
Use publish-only `-dart-publish` tags to support Dart 3.9+ `tag_pattern` releases without mutating the main release tag or branch history.

## Why
Dart `tag_pattern` resolution requires the matched tagged commit to contain a matching `version:` in `pubspec.yaml`.

FishyJoes release tags do not guarantee that, because the Dart version is injected during packaging. This PR creates a separate Dart publish tag whose commit contains the baked Dart version.

## What changed
- Dart flexible-version git dependencies now emit:
  - `tag_pattern: "{{version}}-dart-publish"`
  - a semver `version`
- Release workflows create a detached Dart publish commit/tag with the baked `pubspec.yaml` version.
- Only `release` events push the publish tag and publish artifacts/packages.
- `setup.dart` supports both old `ref` lockfiles and new `tag_pattern` lockfiles.
- Updated the Dart workflow template and regenerated TestAPI workflows.

## How it works
For a release tag like `1.2.3`:

1. The release workflow checks out tag `1.2.3`.
2. CI writes `version: "1.2.3"` into the Dart package pubspec.
3. CI creates a detached commit containing that publish-specific change.
4. CI tags that commit as `1.2.3-dart-publish`.
5. The release workflow pushes the publish tag and publishes from the same run.

Consumers using flexible Dart versions resolve against the publish-only tags via:

```yaml
tag_pattern: "{{version}}-dart-publish"
```

### DIFF
--- a/.github/workflows/iota-runtime.yaml
+++ b/.github/workflows/iota-runtime.yaml
@@ -190,15 +190,68 @@ jobs:
 
       - name: Determine version
         shell: zsh {0}
+        env:
+          RELEASE_TAG: ${{ github.event.release.tag_name }}
         run: |
           set -euxo pipefail
 
-          VERSION=${GITHUB_REF#refs/*/}
-          if [[ ! $VERSION =~ ^[0-9]+\.[0-9]+\.[0-9]+(-[a-zA-Z0-9]+)*$ ]]; then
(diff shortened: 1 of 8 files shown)


cobbal (approved):
lgtm

PR #338, opened 2026-04-21
add wasm strip option

cschear:
Add option to run wasm-strip on generated wasm files to remove debug sections.


cobbal:
Dupe of #339 because I didn't see this.

PR #340, opened 2026-04-27
fix: generate setup code only for mentioned dependency types

cobbal:


mstokercricut (approved):
Seems like a prudent strategy to check both.

PR #357, opened 2026-06-04
Add Python generation support for FishyJoes bindings

antoniogermano:
Adds Python as a FishyJoes-generated binding target using the shared Iota ABI. This includes the Python translator, generated package templates, shared Python runtime package, wheel build support, generated TestAPI and EmojiFun Python bindings, CI workflow templates, and documentation/ADRs for the Python architecture.

Key areas:
- Generates Python wrappers, `.pyi` stubs, `py.typed` (WIP), CFFI declarations, and package metadata.
- Adds `fishyjoes-runtime` for shared Python-side Iota runtime behavior.
- Supports Python snake_case generated from Swift's CamelCase
- Supports Python wheel building and native-library packaging.
- Adds TestAPI Python integration coverage for primitives, strings, collections, structs, enums, references, protocols, async, diagnostics, typing metadata, and packaging.
- Adds Python package naming/configuration support for generated bindings.
- Updates shared generator/runtime behavior needed by Python while preserving existing targets.
- Includes custom warnings on misplaced export annotations (can be split later)

Validation performed:
- `swift test --disable-sandbox --filter 'FishyJoesCoreTests|FishyJoesExecuteTests'`
- TestAPI Python `generate build test pack`
- Python TestAPI suite: 108 tests passing
- Further validation will be executed in future commits

Notes:
- This PR is intended for architecture and release-critical review of Python generation, runtime packaging, CI/CD, and generated-output reproducibility.


documentation/decisions/0002-python-target-uses-iota-abi.md
```diff
@@ -0,0 +1,63 @@
+# 0002. Python Target Uses Iota ABI
+
+## Status
+
+Accepted
+
+## Context
+
+FishyJoes already has the Iota ABI for non-JNI language targets. C# and Dart use
```
cobbal: JNI-or-not doesn't have much to do with this. Typescript, for example, also doesn't use the iota interface and uses the node interface (NAPI).

The main difference between languages that can use iota and those that can't is the ffi libraries for those languages.

In both java and ts, a native C API for interacting with java/ts objects is provided, this API is used from swift.

In dart and C#, the ffi library is on the dart/C# side, and calls into the native library functions, inverting control. Since both are designed to call C libraries not aware of the other languages, they can both use the same iota interfaces.

Python has both kinds of ffi libraries. CPython's C extensions fall into the JNI/NAPI category, since it's a C library for accessing python objects. The ctypes and cffi libraries fall into the iota category, since the ffi is driven from the python side.

documentation/decisions/0002-python-target-uses-iota-abi.md
```diff
@@ -0,0 +1,63 @@
...
+- Copy Kotlin's JNI architecture.
+  This would provide a known language target to imitate, but it would make
+  Python a second non-Iota implementation and duplicate the FFI surface that
+  Iota exists to centralize.
+- Create Python-specific Swift `@_cdecl` exports.
+  This could optimize individual Python cases, but it would fork the ABI and
+  make every exported Swift feature an additional N-language maintenance burden.
+- Build a brand-new shared ABI.
+  This may be attractive long term, but it is larger than adding Python and
+  would risk destabilizing existing C# and Dart users.
```
cobbal: I'm not sure if these are real options. What would be different about the interface needed than is already present in iota?

documentation/decisions/0003-python-bindings-use-generated-cffi-declarations.md
```diff
@@ -0,0 +1,66 @@
...
+generated Swift symbols.
+
+## Options Considered
+
+- Use cffi ABI mode with generated declarations.
+  This gives typed calls and callbacks without compiling Python extension code
+  for the normal path, and it ties Python declarations to FishyJoes codegen.
+- Use `ctypes`.
+  This avoids a third-party dependency, but callback signatures and structured
+  declarations are harder to manage safely at the scale of generated bindings.
```
cobbal, line 23: We control both sides of the generation in ctypes, exactly the same for cffi. The only difference between the two in maintaining the types is whether we generate C interfaces in .h files, or ctypes type lists in .py files.

documentation/decisions/0003-python-bindings-use-generated-cffi-declarations.md
```diff
@@ -0,0 +1,66 @@
...
+- Use a compiled C extension for the whole binding layer.
+  This can be faster and can hide some pointer details, but it creates a larger
+  native code surface and makes packaging/debugging more expensive.
```
cobbal: One of the most significant drawbacks to this approach is that it ties a binary artifact to a particular version of CPython and loses implementation portability.

documentation/decisions/0003-python-bindings-use-generated-cffi-declarations.md
```diff
@@ -0,0 +1,66 @@
...
+- Use a generic dynamic invocation helper.
+  This reduces generated wrapper code, but it moves symbol mistakes to runtime,
+  loses per-function type clarity, and makes code review less effective.
+- Maintain declarations by hand in the Python runtime.
+  This is initially simple, but it will drift from generated Swift exports.
```
cobbal: These two are a different category of thing, neither would be able to create the bridge between native and python code.

documentation/decisions/0007-python-bindings-ship-as-platform-wheels.md
```diff
@@ -0,0 +1,154 @@
...
+and native Swift dynamic libraries. A released package must let Python users
+install the binding with normal Python packaging tools while still getting the
+correct native libraries for their operating system and architecture.
+
+FishyJoes already packages native libraries for C# and Dart on macOS, Linux, and
+Windows. Python should follow the same native-library model rather than invent a
+custom installer.
+
+Python 3.10 is scheduled to reach end of life in 2026. The Python target should
+not launch on a version that is about to fall out of support.
```
cobbal, line 19: Not sure why that should be a goal

documentation/decisions/0008-python-target-uses-red-green-integration-tdd.md
cobbal, line 1: This file feels more like a description of the dev process used for this one PR, not a description of the architecture. Is it still needed?

documentation/decisions/0008-python-target-uses-red-green-integration-tdd.md
```diff
@@ -0,0 +1,149 @@
...
+For example, explicitly passing an enum argument proves enum conversion works,
+but it does not prove that a Swift default such as `join: Join = .bevel`
+translates into idiomatic generated Python. The bypassed behavior remains open
+work until it is fixed or explicitly recorded as deferred.
+
+The refactor gate may make some slices larger than the first failing assertion
+suggests. That cost is acceptable when the alternative is accumulating
+generator/runtime debt that would make later slices harder or less coherent.
+
+Crash summaries omit most of the raw `.ips` detail by design. If a summary is
```
cobbal: I have no idea what an ips file is and what it has to do with FishyJoes architecture.

documentation/decisions/0009-python-cffi-declarations-define-the-abi-contract.md
```diff
@@ -0,0 +1,74 @@
...
+  This is easy for runtime functions, but it cannot safely track generated
+  per-module symbols.
+- Generate one `_declarations.h` file from the same symbol model that emits
+  Swift Iota shims.
+  This makes the generated header the reviewed ABI surface and lets tests verify
+  declarations against actual exported symbols.
+
+## Decision
+
+FishyJoes generates `_declarations.h` as the cffi ABI contract for each Python
```
cobbal: It's unfortunate to bring in a 3rd language (C) that we aren't using to define this interface. This is one of the large drawbacks of cffi

documentation/decisions/0009-python-cffi-declarations-define-the-abi-contract.md
```diff
@@ -0,0 +1,74 @@
...
+## Tradeoffs
+
+Generating a complete ABI header adds generator work before many user-visible
+Python wrappers can run.
+
+Ownership comments are not machine-enforced by C, so the Python runtime still
+needs focused ownership tests.
+
+Declaring ABI details explicitly may expose existing Iota inconsistencies that
+C# and Dart tolerate through different FFI tooling.
```
cobbal, line 75: It shouldn't. The swift-iota interface is the source of truth. The header merely reflects that.

documentation/decisions/0011-python-packages-version-and-load-native-dependencies-explicitly.md
```diff
@@ -0,0 +1,100 @@
...
+  packaged native dependencies explicitly.
+  This is more work, but it makes compatibility and diagnostics much clearer.
+
+## Decision
+
+Each generated Python package declares:
+
+- Its own package version, normally derived from the Swift package version.
+- The FishyJoes runtime package version range it supports.
+- The Iota ABI version it was generated for.
```
cobbal: iota ABI is inextricably linked to the FishyJoes runtime. That's the version number that matters for it.

cobbal (commented):
Still looking through this. Switching to posting comments as I go instead of batching them.

integration-tests/TestAPI/bindings/c-sharp/Cricut.TestAPI.Tests/GCFixture.cs
```diff
@@ -14,9 +14,10 @@ public void Dispose() {
             Console.WriteLine("TEARDOWN");
             Cricut.FishyJoesRuntime.Utilities.PrintOutstandingHandles();
             #if DEBUG
-            if (Cricut.FishyJoesRuntime.Utilities.OutstandingHandleCount() != 2) {
+            var outstandingHandleCount = Cricut.FishyJoesRuntime.Utilities.OutstandingHandleCount();
+            if (outstandingHandleCount != 0 && outstandingHandleCount != 2) {
```
cobbal, line 23: Why are you expecting the handles for true and false to disappear?

integration-tests/TestAPI/Sources/TestAPI/DefaultArguments.swift
```diff
@@ -6,4 +6,19 @@ public enum DefaultArguments {
     public static func echoDefaults(y: Int? = nil, x: Int?, theLabelForZ z: Double = 3.14) -> String {
         "\(x ?? 42) \(y ?? 52) \(z)"
     }
+
+    /// <!-- FishyJoes.export(describeEnumDefault) -->
+    public static func describeEnumDefault(_ value: Int, color: SimpleEnum = .blue) -> String {
```
cobbal, line 11: This will have a FishyJoes warning in the generate step. Which is fine, since it might be an important code corner to test, but a comment should mention that.

integration-tests/TestAPI/Sources/TestAPI/Protocols.swift
```diff
@@ -51,6 +51,21 @@ public struct AProtocolImplementation: AProtocol {
     }
 }
 
+/// <!-- FishyJoes.export(ProtocolWitnesses) -->
+public enum ProtocolWitnesses {
```
cobbal: This should be named something different. A protocol witness is a technical term used elsewhere in the project to refer to the type-erased record of protocol methods. Maybe this should be named AProtocolUser

integration-tests/TestAPI/bindings/.gitignore
```diff
@@ -21,3 +21,5 @@ package-lock.json
 DebugGenerated
 .idea
 *.DotSettings.user
+/python/.venv/
+/python/**/__pycache__/
```
cobbal: Since there's also a .gitignore in the python directory, these new lines should probably go there

python-runtime/_build_wheel.py
cobbal, line 1: This seems like a complicated file. Where did it come from? Is it generated by something or hand-written?

README.md
```diff
@@ -86,6 +93,13 @@ public struct SomeSwiftType {
     /// <!-- FishyJoes.export(abcMethod, cSharp: ABCMethod) -->
     public func abcMethod() { /* ... */ }
 }
+
+// Invalid: this annotation is inside the method body, so it is not attached to
+// the declaration and FishyJoes will warn that it is ignored.
+public func misplacedExportExample() -> Int {
+    /// <!-- FishyJoes.export(misplacedExportExample) -->
+    return 1
+}
```
cobbal: The main readme should try to contain relevant things to most users. The annotation on the wrong line thing was as far as I know 1 typo in 1 downstream repository, and probably doesn't deserve a mention here.

Sources/FishyJoesExecute/Phases/DartPhases.swift
```diff
@@ -28,6 +28,14 @@ class DartPhases: IotaPhases, Phases {
...
                 ]
             }
+            if let editedPath = options.editedDependencyPaths[dependency.identity.lowercased()] {
+                let dependencyPath = relativePath(of: editedPath, relativeTo: "bindings/dart/generated")
+                return lines + [
+                    "  path: \(dependencyPath)/\(depNames.path)",
+                ]
+            }
+            // Always use exact version for git ref: git does not support semver range syntax.
+            // Flexible versions for Dart are expressed via flutter-package.json (npm), not pubspec.yaml.
```
cobbal: This is no longer true in main

Sources/FishyJoesExecute/Phases/DartPhases.swift
```diff
@@ -28,6 +28,14 @@ class DartPhases: IotaPhases, Phases {
                     "  path: DEPENDENCY_NOT_FOUND",
                 ]
             }
+            if let editedPath = options.editedDependencyPaths[dependency.identity.lowercased()] {
+                let dependencyPath = relativePath(of: editedPath, relativeTo: "bindings/dart/generated")
+                return lines + [
+                    "  path: \(dependencyPath)/\(depNames.path)",
+                ]
+            }
```
cobbal, line 40: Why was this changed?

Sources/FishyJoesExecute/Phases/DartPhases.swift
```diff
@@ -74,15 +129,22 @@ class DartPhases: IotaPhases, Phases {
         return [
             "__PUBSPEC_DART_DEPENDENCIES__": join(lines: pureDartDependencyLines, indent: 2),
             "__PUBSPEC_FLUTTER_DEPENDENCIES__": join(lines: flutterDependencyLines, indent: 2),
+            "__PUBSPEC_DART_DEPENDENCY_OVERRIDES__": dependencyOverrides(pureDartDependencyOverrideLines),
+            "__PUBSPEC_FLUTTER_DEPENDENCY_OVERRIDES__": dependencyOverrides(flutterDependencyOverrideLines),
```
cobbal: overrides don't seem appropriate. The library is generated with a local dependency. That's not an override

Sources/FishyJoesExecute/Phases/DartPhases.swift
```diff
@@ -100,12 +162,17 @@ class DartPhases: IotaPhases, Phases {
     func testPhase() throws {
         // Use dart to execute the test suite
         try withDirectory("bindings/dart/generated") {
-            let env = options.codeCoveragePath.map {
-                [
-                    "LLVM_PROFILE_FILE": "\($0)/fishy-joes-test-\(platform)-\(UUID()).profraw",
-                ]
-            } ?? [:]
-            try cmd("dart", "test", "--chain-stack-traces", addEnv: env).run()
+            let tempDirectory = "\(FileManager.default.currentDirectoryPath)/.dart_tool/fishyjoes-test-tmp"
```
cobbal, line 199: Why is this done?

Sources/FishyJoesExecute/Phases/KotlinPhases.swift
```diff
@@ -29,10 +56,10 @@ class KotlinPhases: BasePhases, Phases {
...
 
     func installPhase() throws {
-        // Install the module library and interfacing JNI library
-        try installLibrary(options.config.module)
-        try installLibrary("\(options.config.module)-java")
-        try options.config.extraDynamicLibraries.forEach { try installLibrary($0) }
+        let translatedLibraries = ([options.config.module] + options.config.requiredModules)
+            .flatMap { [$0, "\($0)-java"] }
+        let nativeLibraries = options.config.extraDynamicLibraries + translatedLibraries
+        try nativeLibraries.forEach { try installLibrary($0) }
```
cobbal, line 66: Dependency dylibs are not part of the downstream package and should not be installed

Sources/FishyJoesIotaRuntime/IotaConverter.swift
```diff
@@ -486,6 +518,13 @@ extension String: IotaConverter {
     }
 
     public static func toIota(_ value: Self, env: Env) throws -> foreignObject {
+        if utf8Constructor.isInitialized(env) {
+            let bytes = Array(value.utf8CString)
```
cobbal: No point in using utf8CString since the null terminator isn't used, just use utf8

Sources/FishyJoesIotaRuntime/IotaConverter.swift
```diff
@@ -468,14 +468,46 @@ public func String_iota_setup(
...
 struct AllocationError: Error {}
+struct StringDecodingError: Error {}
 
 extension String: IotaConverter {
     fileprivate static var getLengthMethod = Env.CallbackMap<(@convention(c) (foreignObject, foreignOutExn) -> Int)>()
     fileprivate static var getUtf16Method = Env.CallbackMap<(@convention(c) (foreignObject, UnsafeMutablePointer<unichar>, foreignOutExn) -> Void)>()
     fileprivate static var constructor = Env.CallbackMap<(@convention(c) (UnsafePointer<unichar>, Int, foreignOutExn) -> foreignObject)>()
+    fileprivate static var getUtf8LengthMethod = Env.CallbackMap<(@convention(c) (foreignObject, foreignOutExn) -> Int)>()
+    fileprivate static var getUtf8Method = Env.CallbackMap<(@convention(c) (foreignObject, UnsafeMutablePointer<CChar>, foreignOutExn) -> Void)>()
+    fileprivate static var utf8Constructor = Env.CallbackMap<(@convention(c) (UnsafePointer<CChar>, Int, foreignOutExn) -> foreignObject)>()
```
cobbal: Since we expect only one of these sets of callbacks to be used, it should instead be stored as an enum of the 2 cases.

Sources/FishyJoesNodeRuntime/TypeSetup.swift
```diff
@@ -8,6 +8,7 @@ public func registerModuleRuntime(env: NAPI.Env, exports: NAPI.Value) throws ->
     try env.setNamedProperty(exports, "default", module)
 
     try RuntimeModule.nodeSetup(env: env, module: module)
+    try installNodeCleanup(env: env, module: module)
```
cobbal: This isn't really a safe function to have exposed publicly. The library isn't expected to ever stop once started, and exit hooks to perform actual cleanup are better left to the OS

Sources/ToolchainConfig/tool-versions.json
```diff
@@ -32,5 +32,10 @@
         "ndkArchName": "x86_64"
       }
     ]
+  },
+  "linuxContainer": {
+    "//": "Used by generated Linux Python workflow jobs.",
+    "image": "ghcr.io/cricut/android-swift-runtime",
+    "version": "4.0.2"
```
cobbal, line 39: This has been removed in main

Sources/FishyJoesExecute/InstallSystemDependencies.swift
```diff
@@ -315,8 +317,14 @@ struct InstallSystemDependencies: ParsableCommand {
             #elseif os(Linux)
             let tempDir = try cmd("mktemp", "-d").runString()
             defer { try? cmd("rm", "-rf", tempDir).run() }
+            try installMintBuildDependenciesOnLinux()
             try cmd("git", "clone", "--depth=1", "https://github.com/yonaskolb/Mint.git", "\(tempDir)/mint").run()
-            try cmd("swift", "run", "--package-path=\(tempDir)/mint", "mint", "install", "yonaskolb/mint").run()
+            try cmd("swift", "build", "--package-path=\(tempDir)/mint", "--configuration", "release", "--product", "mint").run()
+            try cmd("install", "-m", "0755", "\(tempDir)/mint/.build/release/mint", "/usr/local/bin/mint").run()
```
cobbal: We should not be modifying ad-hoc root paths like this on random linux distributions.

Sources/FishyJoesExecute/InstallSystemDependencies.swift
```diff
@@ -337,4 +345,132 @@ struct InstallSystemDependencies: ParsableCommand {
             #endif
         }
     }
+
+    func installMintBuildDependenciesOnLinux() throws {
+        var packages = [String]()
+        if !cmd("test", "-f", "/usr/include/sqlite3.h").runBool() {
```
cobbal: Hard coding paths should be avoided here too

Sources/FishyJoesExecute/Resources/bindings-template/python/__TEMPLATE__/pyproject.toml
```diff
@@ -0,0 +1,19 @@
+# __AUTOGEN_WARNING__
+
+[project]
+name = "__PYTHON_DISTRIBUTION_NAME__"
+version = "__PYTHON_PACKAGE_VERSION__"
+description = "Python bindings generated by FishyJoes for __MODULE_NAME__"
+requires-python = ">=3.11"
+dependencies = [
+  "__PYTHON_RUNTIME_DISTRIBUTION_NAME__>=0.0.1",
```
cobbal: version hardcoded, should probably just be part of `__PYTHON_DEPENDENCIES__`

Sources/FishyJoesExecute/Resources/bindings-template/python/__TEMPLATE__/src/__PYTHON_IMPORT_PACKAGE__/_diagnostics.py
```diff
@@ -0,0 +1,9 @@
+# __AUTOGEN_WARNING__
+
+from __future__ import annotations
+
+from . import _native
+
+
+def diagnostics():
```
cobbal: If this is a public thing it should be documented. I have no idea what it's supposed to do yet

Sources/FishyJoesExecute/Resources/bindings-template/python/tests/test_example.py
```diff
@@ -0,0 +1,9 @@
+import unittest
+
+import __PYTHON_IMPORT_PACKAGE__ as bindings
```
cobbal: Since this file is setting the example, naming is important. Use the library name instead of "bindings"

Sources/FishyJoesExecute/SwiftPMWorkspaceState.swift
```diff
@@ -0,0 +1,46 @@
...
+
+        let packageRef: PackageRef
+        let state: State
+    }
+
+    let object: Object
+
+    static var current: SwiftPMWorkspaceState {
+        guard
+            let data = try? Data(contentsOf: URL(fileURLWithPath: ".build/workspace-state.json")),
```
cobbal, line 41: This file seems almost completely undocumented. Should we be using it?

Sources/FishyJoesExecute/SwiftPackage.swift
```diff
@@ -281,6 +281,33 @@ extension SwiftPackage.Dependency {
         }
     }
 
+    func versionInPythonRequirementFormat(flexibleVersions: Bool = false) -> String? {
+        switch self {
+        case .sourceControl(_, _, .branch), .sourceControl(_, _, .revision), .fileSystem:
+            return nil
```
cobbal, line 287: For file-local, should we not be using one of the local forms documented here? https://pip.pypa.io/en/stable/reference/requirements-file-format/

Sources/FishyJoesCore/Unparse/CSharpClass.swift
```diff
@@ -103,8 +103,20 @@ class CSharpClass: NestedClass {
...
     }
 
+    func memberIdentifier(_ name: String) -> String {
+        var identifier = CSharpClass.deforbidify(name)
+        let nestedTypeNames = Set(innerClasses.map(\.unqualifiedName))
+        if nestedTypeNames.contains(identifier) {
+            identifier += "_"
+        }
+        return identifier
+    }
```
cobbal, line 120: I don't think we should be doing this automatically, as it can lead to weird, unstable interfaces where adding a type suddenly causes  possibly-unrelated field to get renamed. Instead, let it be a compiler error that the library dev can go and export to a different name manually

Sources/FishyJoesCore/Unparse/CSharpClass.swift
```diff
@@ -103,8 +103,20 @@ class CSharpClass: NestedClass {
...
+
     func output(field: Variable, to fragment: SourceFragment) {
         document(field.documentation, fragment: fragment)
+        let propertyName = memberIdentifier(field.name)
+        let getterMethodName = memberIdentifier("Get\(CSharpClass.deforbidify(field.name))")
```
cobbal, line 125: No need to call deforbidify here, There's not much point to avoiding conflicts if we're going to be appending a string to it anyway

Sources/FishyJoesExecute/CodeGen.swift
```diff
@@ -81,16 +85,19 @@ public class CodeGen: ParsableCommand {
...
 
     lazy var buildConfig: BuildConfiguration = {
         // the "module-bindings" subdirectory is needed to avoid this: https://stackoverflow.com/a/71759561
         let swiftBindingsRoot = "bindings/swift-interfaces/generated/\(config.module)-bindings"
 
         // Assemble a build configuration from passed arguments
-        let localPathsNeeded = packageInfo.dependencyMap.entries.values.map(\.localPath)
+        let localPathsNeeded = packageInfo.dependencyMap.entries.values.map { localPath(for: $0) }
 
+        let preferResolvedLocalDependencies = ProcessInfo.processInfo.environment["FISHYJOES"] == "1"
```
cobbal: This is a very odd test here, because this environment variable is used by other packages to determine if FIshyJoes should be a dependency. Since we're already inside the code for FishyJoes, this doesn't make sense to me

python-runtime/src/fishyjoes_runtime/config.py
```diff
@@ -0,0 +1,138 @@
...
+    package_dir: Path
+    module_name: str
+    package_version: str
+    register_types_symbol: str
+    native_dir_candidates: Sequence[Path]
+    runtime_distribution_name: str = "fishyjoes-runtime"
+    dependencies: Sequence[RuntimeDependency] = ()
+    declaration_files: Sequence[str] = ("_declarations.h", "_generated_declarations.h")
+    iota_abi_version: str = IOTA_ABI_VERSION
+    runtime_requirement: str = ">=0.0.1"
```
cobbal, line 53: Is this code meant for runtime or package-time? This syntax looks like package-time syntax

python-runtime/src/fishyjoes_runtime/config.py
```diff
@@ -0,0 +1,138 @@
...
+    if not isinstance(data, dict):
+        raise RuntimeError(f"Invalid FishyJoes wheel metadata: {metadata_path}")
+    return data
+
+
+def validate_platform_compatibility(config: RuntimeConfig) -> None:
+    current_platform = platform.system()
+    if current_platform not in config.supported_platforms:
+        supported = ", ".join(config.supported_platforms)
+        raise RuntimeError(f"Unsupported Python binding platform: {current_platform}; supported platforms: {supported}")
```
cobbal, line 119: I think a lot of this function runs into "easier to ask forgiveness than permission". We should just try to load the library instead of checking an entire matrix of what we expect

python-runtime/src/fishyjoes_runtime/dependencies.py
```diff
@@ -0,0 +1,38 @@
...
+        raise RuntimeError(
+            f"Dependency package {dependency.distribution_name} Iota ABI version mismatch: "
+            f"package requires {dependency.iota_abi_version}, dependency provides {dependency_iota_abi}"
+        )
+    if not report.get("supported", False):
+        raise RuntimeError(f"Dependency package {dependency.distribution_name} did not finish setup")
+    return report
+
+
+def load_dependencies(dependencies) -> dict[str, dict[str, object]]:
```
cobbal: missing a type annotation for dependencies

python-runtime/src/fishyjoes_runtime/native.py
```diff
@@ -0,0 +1,133 @@
...
+from collections.abc import Sequence
+from dataclasses import dataclass
+from pathlib import Path
+
+
+@dataclass(frozen=True)
+class NativeLibrary:
+    name: str
+    path: Path
+    reused: bool = False
```
cobbal, line 23: This is not obvious what it means, it should be documented

python-runtime/src/fishyjoes_runtime/native.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+def read_declarations(package_dir: Path, declaration_files: Sequence[str]) -> str:
+    declarations: list[str] = []
+    for declaration_file in declaration_files:
+        path = package_dir / declaration_file
+        if path.exists():
+            declarations.append(path.read_text(encoding="utf-8"))
+        elif declaration_file == "_declarations.h":
+            raise RuntimeError(f"Missing required Iota declarations file: {path}")
```
cobbal, line 177: This is really odd. Why does it only raise errors when filenames are specific things? Feels like hiding errors.

python-runtime/src/fishyjoes_runtime/native.py
```diff
@@ -0,0 +1,133 @@
...
+    if system == "Windows":
+        return f"{name}.dll"
+    raise RuntimeError(f"Unsupported Python binding platform: {system}")
+
+
+def runtime_native_dir_candidates() -> list[Path]:
+    return [_RUNTIME_PACKAGE_DIR / "native"]
+
+
+def library_path(name: str, native_dir_candidates: Sequence[Path], build_hint: str) -> Path:
```
cobbal, line 78: I think we should always know where to find the library. Searching for it in more than one place may give more opportunities for stale, mismatched versions of the library to be found instead.

python-runtime/src/fishyjoes_runtime/iota.py
cobbal, line 1: There are a lot of types in here that don't have a corresponding typed interface. e.g. SwiftRange, AttributedString, many others have no entry in the pyi file

python-runtime/src/fishyjoes_runtime/_declarations.h
cobbal, line 1: Most of this header is declaring functions present in the FishyJoesIotaRuntime. I would expect it to instead be part of the /python-runtime library, next to the iota.py that references it

Sources/FishyJoesCore/TranslatedTypes/TranslatedReference.swift
cobbal, line 1: I expected more code in this directory. It looks like PythonTranslator is instead switching on the dynamic subtypes of TranslatedType, which can lead to fragile code and forgotten cases.

integration-tests/TestAPI/bindings/c-sharp/Cricut.TestAPI.Tests/GCFixture.cs
antoniogermano: The true/false handles only exist once the runtime's lazy Loader initialization runs; `Swift_Bool_setup` registers them on the first actual Swift call (since they are lazy).

Places such as `Utilities.TrackAllocationsDebug`, `OutstandingHandleCount`, `PrintOutstandingHandles` live in the standalone `Utilities` class and never touch `Loader`. So any session whose executed tests never go through Swift (e.g. `dotnet test --filter` subsets during debugging) tears down with a legitimate 0.

This bit me while fixing the naming-collision regressions (tests were failing), so I made the change.

A real leak still fails (any count other than 0 or 2).

integration-tests/TestAPI/Sources/TestAPI/DefaultArguments.swift
antoniogermano: Agreed, it's intentional warning coverage. Adding a comment saying the generate-step warning is expected

integration-tests/TestAPI/Sources/TestAPI/Protocols.swift
antoniogermano: Agreed; what about `ProtocolFixtures`? Let me know and I'll rename it.

integration-tests/TestAPI/bindings/.gitignore
antoniogermano: Moved into `bindings/python/.gitignore`

python-runtime/_build_wheel.py
antoniogermano: This is not generated, but I wrote it with the help of Claude. It ended up with three jobs:
- pyproject metadata + version stamping
- native library collection (the `_pe_*` functions parse PE imports to find Swift runtime DLLs on Windows)
- per-platform wheel repair

I can break it down a bit, adding a module docstring describing the three phases, and extract the Windows PE-parsing helpers into its own module (like `_windows_pe.py`) so the flow is readable.

I'll work on it

Sources/FishyJoesExecute/Phases/DartPhases.swift
antoniogermano: `localPath(for:)` checks SwiftPM edited dependencies first, then falls back to FishyJoes’ existing local path for the dependency: the filesystem dependency path for local dependencies, or `.build/checkouts/<repo>` for remote dependencies.

The goal is to make local generation/build steps use a SwiftPM-edited dependency checkout when one exists, instead of the resolved checkout.

In `DartPhases`, edited dependencies are handled before the versioned git dependency path, so they become local `path:` dependencies for development. Non-edited remote dependencies keep the same behavior as `main`: `tag_pattern` + `version` when flexible versions are enabled, otherwise an exact `ref`.

Sources/FishyJoesExecute/Phases/DartPhases.swift
antoniogermano: True, I had actually reworked the code based on main and forgot to update the comment. Done

Sources/FishyJoesExecute/Phases/DartPhases.swift
antoniogermano: Agreed for the direct dependency. If the generated package already has a local `path:` dependency, emitting the same package under `dependency_overrides` is redundant and confusing.

The case I was trying to cover was local development resolution across a dependency graph. For example:
```
my_generated_app
  depends on fishyjoes_dart via local path
  depends on some_package via local path

some_package
  depends on fishyjoes_dart via git ref 7.0.0
```
Without an override, the root package and `some_package` can ask for `fishyjoes_dart` from different sources. With an override at the root, Dart is told: “for this whole resolution, use the local `fishyjoes_dart`”.

So I’ll separate those concerns: `pubspec.yaml` should describe the direct dependency, and local-development graph forcing should move to `pubspec_overrides.yaml` instead of being emitted into the generated publishable pubspec.

When working on Python I found a few things that I thought would make development easier (and which I relied on during development), that's one of those changes.

Sources/FishyJoesExecute/Phases/DartPhases.swift
antoniogermano: Dart tests, especially native/FFI tests, can create temp/build/runtime files. Running many tests in parallel and using the shared host temp directory can make CI flaky or too resource-heavy. This forces temp output into a package-local directory and reduces parallelism.

I’ll add a comment here because the reason is not obvious from the code.

documentation/decisions/0002-python-target-uses-iota-abi.md
antoniogermano: Agreed, reworded. The ADR now describes whether Swift drives a host-runtime API (like JNI/NAPI) or the host language drives a C ABI FFI against the shared Iota model (like C#/Dart/Python cffi)

documentation/decisions/0002-python-target-uses-iota-abi.md
antoniogermano: I think you're right, those weren't meaningful alternatives. I dropped them from the ADR

documentation/decisions/0003-python-bindings-use-generated-cffi-declarations.md
antoniogermano: The bullet's "harder to manage safely" wording should be about the ergonomics/reviewability of generated declarations (C declarations in .h versus ctypes type lists in .py), not about ownership. But I think it can be read as a control asymmetry... would you like me to reword it?

documentation/decisions/0003-python-bindings-use-generated-cffi-declarations.md
antoniogermano: Good point — added to the ADR

documentation/decisions/0003-python-bindings-use-generated-cffi-declarations.md
antoniogermano: Right, they're an invocation strategy after an FFI bridge exists. They're no longer listed as FFI options

documentation/decisions/0007-python-bindings-ship-as-platform-wheels.md
antoniogermano: I added this because I wanted to avoid launching new bindings on a CPython version that's about to hit upstream EOL (writing new code that quickly becomes deprecated, while potentially preventing us from using new Python features just to keep compatibility with something that's already getting deprecated), so I set the floor at the oldest actively supported version

documentation/decisions/0008-python-target-uses-red-green-integration-tdd.md
antoniogermano: Right, this might not actually be an ADR but a dev process; I figured it would be safest to create the Python bindings by porting existing tests first then using them as baseline/grounding for implementation, but it's not needed as an ADR so I deleted it

documentation/decisions/0008-python-target-uses-red-green-integration-tdd.md
antoniogermano: That was definitely for the development/debugging process; removed ADR 0008 entirely

documentation/decisions/0009-python-cffi-declarations-define-the-abi-contract.md
antoniogermano: Added to the tradeoffs: cffi ABI mode brings a third language (C) into a boundary that's otherwise Swift and Python

documentation/decisions/0009-python-cffi-declarations-define-the-abi-contract.md
antoniogermano: Updated ADR, I'll also create a follow up improvement on our ticketing tool

documentation/decisions/0011-python-packages-version-and-load-native-dependencies-explicitly.md
antoniogermano: Removed the separate public ABI version: packages no longer declare or check it, and the runtime keeps only an internal schema number that diagnostics() reports for debugging. The runtime version requirement is the compatibility gate.

integration-tests/TestAPI/Sources/TestAPI/DefaultArguments.swift
antoniogermano:
```
  // Intentionally exercises a default the generator cannot translate for
  // every language; the generate step is expected to emit a warning here.
  /// <!-- FishyJoes.export(echoDefaultTolerance) -->
```

integration-tests/TestAPI/Sources/TestAPI/Protocols.swift
antoniogermano: Renamed to `ProtocolFixtures`, though we can always change it to something else

python-runtime/_build_wheel.py
antoniogermano: I added the module-level phase description and moved the Windows PE import parsing into `_windows_pe.py` so `_build_wheel.py` is responsible for metadata stamping, native collection, and repair orchestration.

README.md
antoniogermano: IIRC I found at least 3 of these, but agreed that it doesn't need to be mentioned here. Removed

Sources/FishyJoesExecute/Phases/DartPhases.swift
antoniogermano: Publishable pubspecs no longer carry `dependency_overrides`; local-development graph forcing now goes to pub's dedicated `pubspec_overrides.yaml` (gitignored, removed when no local deps are in play) before build/test and when assembling the flutter package)

Sources/FishyJoesExecute/Phases/DartPhases.swift
antoniogermano: Comment added

Sources/FishyJoesExecute/Phases/KotlinPhases.swift
antoniogermano: Agreed. `installPhase` now installs only the module's own libraries plus explicit `extraDynamicLibraries`; `requiredModules` are still built for local tests but their dylibs come from their own packages (I think ideally Gradle included builds locally, Maven artifacts when published).

Sources/FishyJoesIotaRuntime/IotaConverter.swift
antoniogermano: Implemented on `2a0c77dc`: now uses `Array(value.utf8)` with the exact byte count; adds a regression test asserting the constructor receives exactly the UTF-8 bytes and count across 1-4 byte sequences and a large payload.

Sources/FishyJoesIotaRuntime/IotaConverter.swift
antoniogermano: Done: the string callbacks are now stored as a single two-case enum per environment, so the one-codec invariant is a type rather than a convention. Setup symbols and signatures are unchanged; UTF-8 still takes precedence when both are registered

Sources/FishyJoesExecute/InstallSystemDependencies.swift
antoniogermano: Not sure if this change is enough, but tool installs now resolve to a user-writable `~/.local/bin` instead of `/usr/local/bin`; generation build-dependency installation was split out of the generic installer.

Sources/FishyJoesExecute/InstallSystemDependencies.swift
antoniogermano: Agreed. The hardcoded `/usr/include/sqlite3.h` check and the `ldconfig` substring probe are gone with the install rework. That also removes the ncurses false-positive where `libncurses.so.6` made us skip the dev package the linker needs.

Sources/FishyJoesExecute/Resources/bindings-template/python/__TEMPLATE__/pyproject.toml
antoniogermano: The full runtime requirement is templated now (distribution name and requirement are both placeholders), derived from the current FishyJoes runtime version with bounded compatible-release pinning instead of a hardcoded >=0.0.1

Sources/FishyJoesExecute/Resources/bindings-template/python/__TEMPLATE__/src/__PYTHON_IMPORT_PACKAGE__/_diagnostics.py
antoniogermano: Documented: it returns package and version, whether native setup finished, Python/platform, runtime package/version/requirement, per-dependency reports, loaded native-library paths.

It's intended to be the first thing to check when an import fails

Sources/FishyJoesExecute/Resources/bindings-template/python/tests/test_example.py
antoniogermano: Done: the example now imports the generated package by its real name instead of aliasing it to 'bindings'

Sources/FishyJoesExecute/SwiftPMWorkspaceState.swift
antoniogermano: Documented. It reads SwiftPM's workspace-state file only to discover edited-dependency paths (I'm not aware of a supported machine-readable command for that in SwiftPM), decodes only the fields it needs, and any read/decode failure degrades to 'no edited dependencies', so a format change can only disable the convenience, never break a build.

Since this is a development convenience we can probably reconsider or remove this, I'll create a ticket for future improvement

Sources/FishyJoesExecute/SwiftPackage.swift
antoniogermano: Agreed. Filesystem dependencies now use pip's local requirement form. A local-path dependency generates a PEP 508 direct reference (`name @ file:///…/bindings/python/generated`) in dev builds instead of a fake version range; publishable wheels still require concrete distribution/version requirements (`pack` rejects filesystem/branch/revision deps without one), and direct references can't be published to a package index.

Sources/FishyJoesCore/Unparse/CSharpClass.swift
antoniogermano: Removed auto-deduping, which now lead to a fatalError. Renamed a few things on our downstream libraries to prevent conflicts under new behavior. Extended same behavior to Python, however not all of Python's automatic deduping were removed. Marking as future improvement

Sources/FishyJoesCore/Unparse/CSharpClass.swift
antoniogermano: `deforbidify` before the conflict check was intentional so escaped keywords participate in collision detection. Conflicts are now a generation-time error

Sources/FishyJoesExecute/CodeGen.swift
antoniogermano: I see how that was confusing. Preferring resolved local dependency paths now has its own switch, `FISHYJOES_PREFER_RESOLVED_LOCAL_DEPS=1`, set by generated CI workflows alongside `FISHYJOES`; `FISHYJOES` remains the downstream-context signal only.

python-runtime/src/fishyjoes_runtime/config.py
antoniogermano: Runtime, but intentionally calling package-time code. Generated packages record compatibility as ordinary PEP 440 requirement strings and the runtime evaluates them at import so incompatibility fails before any FFI call. It's now documented on the function so it doesn't read as misplaced package-time code

python-runtime/src/fishyjoes_runtime/config.py
antoniogermano: I agree conceptually, though I think declared platform/architecture metadata can produce better errors than raw `dlopen`. Updated so that the validation checks only what the package declares about itself (Python version, runtime version, platform/architecture, wheel metadata), while the actual `dlopen` proves library availability and wraps the real failure with context. That boundary is now stated on the function

python-runtime/src/fishyjoes_runtime/dependencies.py
antoniogermano: Done: annotated as `Sequence[RuntimeDependency]`

python-runtime/src/fishyjoes_runtime/native.py
antoniogermano: Documented: reused means the registry returned an already-loaded library for the name (a cache hit) rather than performing a fresh load

python-runtime/src/fishyjoes_runtime/native.py
antoniogermano: Agreed, that check hid the contract. Every declaration file the config lists is now required (the config names exactly what the package was generated with), and a missing one fails loudly naming the file

python-runtime/src/fishyjoes_runtime/native.py
antoniogermano: Agreed, I made some changes but I think this can still be improved; will open internal ticket.

Now a missing library errors listing every checked path, distinct copies in more than one root raise an explicit 'ambiguous native library' error, and the process-wide registry refuses to load a second copy of an already-loaded library. Installed-wheel and local-build layouts share one documented candidate design with only one root populated in normal operation

python-runtime/src/fishyjoes_runtime/iota.py
antoniogermano: The runtime now has a typed public interface (including SwiftRange and the Iota helper types), generated stubs are validated by `mypy --strict`, `stubtest`, and `pyright --verifytypes` on the reference binding, and the AttributedString family is typed end-to-end.

Sources/FishyJoesCore/TranslatedTypes/TranslatedReference.swift
antoniogermano: Refactored, Python type support is now a requirement on TranslatedType itself (ADR 0015), so each translated type makes an explicit Python decision and new kinds fail at compile time rather than silently missing Python conversion/stub support

python-runtime/src/fishyjoes_runtime/_declarations.h
antoniogermano: Done: the shared `FishyJoesIotaRuntime` declarations moved into `fishyjoes_runtime/_declarations.h` (owned and shipped by the runtime, next to `iota.py`); generated packages contribute only their module-specific declarations (`_generated_declarations.h`).

The runtime loads its own declarations first, then the package's.

PR #359, opened 2026-06-09
Update copy-linux-swift-stdlib.sh

roylanceMichael:
flutter was missing this file for linux:

Launching lib/main.dart on Linux in debug mode...
CMake Error at cmake_install.cmake:412 (file):
  file INSTALL cannot find
  "/home/mroylance/cricut/roylance_research/tooligans/linux/flutter/ephemeral/.plugin_symlinks/fishyjoes_dart/linux/native/libswiftSwiftOnoneSupport.so":
  No such file or directory.

Building Linux application...                                           
Error: Build process failed


cobbal (approved):
Oh yeah, that would be good to have for unoptimized builds.

PR #370, opened 2026-09-18
feat: add CIExtraSetupSteps to insert workflow steps after checkout in generated workflows

bdolewski-intellias:
## Summary

Adds an optional `CIExtraSetupSteps` key to `fishy-joes.yaml`. It takes a YAML sequence of ordinary workflow steps and inserts them right after the checkout step in every generated job, including `regenerate-check` and ahead of the Windows pre-resolve step. Nothing vendor-specific lands in FishyJoes.

```yaml
CIExtraSetupSteps: |
  - name: Log in to the package registry
    run: swift package-registry login https://registry.example.com --token ${{ secrets.REGISTRY_TOKEN }}
```

## Why

A library that declares a dependency as `.package(id: "scope.name")` needs registry auth on every runner that resolves its package graph. The `CIPreBuildHook` cannot cover this. The `regenerate-check` job runs `swift run -- fishy-joes generate` before any hook, and the Windows jobs run `swift package resolve` before the hook too. Only steps that follow checkout cover all of them.

This is the generalized hook the `TODO: generalize this to a more flexible hook system when n>=3` note in `ProjectConfig` asks for. Consumers keep their own login step, whatever registry they use, in their own config.

## Changes

- `ProjectConfig`: new `ciExtraSetupSteps: String?`, parsed from `CIExtraSetupSteps`. The value must parse as a YAML sequence of step mappings, otherwise the config is rejected.
- `FileTemplater`: dedents the steps and appends them to `__CI_GIT_STEPS__` after checkout. Nothing changes for projects that do not set the key.
- `Utilities`: small `dedent(lines:)` helper.
- `PackageInit`: passes `nil` for the new field.
- `FileTemplaterTests`: absent key, rendered steps with relative indentation preserved, and config validation.

## History

The first version of this PR emitted a vendor-specific composite action (`CIRegistryAuth`). Reworked after review feedback that no vendor-specific CI belongs in this repo.

## Not in this PR

- `integration-tests/TestAPI` does not set the key, so its committed generated workflows are unchanged.

## Test

```bash
swift test --filter FileTemplaterTests
```


bdolewski-intellias:
🤖 **Windows: Integration tests fails before any FishyJoes code runs, and it fails the same way on `main`.**

The job dies inside `dart-lang/setup-dart@v1`:

```
##[error]Unable to process command '::add-matcher::D:\a\_actions\compnerd\gha-setup-swift\eeda069c.../dart-analyzer.json' successfully.
##[error]Could not find file 'D:\a\_actions\compnerd\gha-setup-swift\eeda069c...\dart-analyzer.json'.
```

`setup-dart` looks for its problem matcher under the previous action's path (`compnerd/gha-setup-swift`), so `GITHUB_ACTION_PATH` is stale when the matcher is registered. The latest `main` run of Integration Tests (2026-09-08, commit 13480b2e, same commit that passed on 2026-07-15) fails on the identical error: https://github.com/cricut/FishyJoes/actions/runs/34279137020

This PR only touches `FishyJoesExecute` Swift sources and a unit test, so it cannot affect the Dart SDK setup step. I have left the Windows job alone. If you want, I can open a separate PR that pins `setup-dart` to a newer release or moves it ahead of `gha-setup-swift` in the job.

bdolewski-intellias:
🤖 The Windows job fix is up as a separate PR: cricut/FishyJoes#371 (run `setup-dart` before `gha-setup-swift`). Root cause is `setup-dart` 1.8.0's new problem matcher reading a stale `GITHUB_ACTION_PATH` left by the composite Swift action.

bdolewski-intellias:
🤖 **Reworked to be vendor-neutral.** The key is now `CIExtraSetupSteps`: a YAML sequence of ordinary workflow steps inserted right after checkout in every generated job. FishyJoes itself no longer references any specific action, host, or registry scope. A registry login step lives in the consuming project's own `fishy-joes.yaml`. Description updated; tests cover the rendering and the config validation.

jaydenrasmussen (approved):

